PR #25, opened 2021-04-11
adding rest of autumus containers!

vsoch:


vsoch:
The error is a "no space left on device" error, but the majority of builds were successful. Since I tested each locally, I think we are okay to merge. I'm going to see if I can better cleanup during the tests.

PR #30, opened 2021-04-13
adding a bunch of bids and docker core images!

vsoch:


vsoch:
Note that the final test might not pass if we have too many containers and the CI runs out of disk space. But all packages I tested locally.

PR #50, opened 2021-04-18
Update bids/validator to 1.7.0

manbat:
Update bids/validator to 1.7.0


alecbcs:

Issue #359, opened 2021-05-27
Be able to define properties for install


vsoch:
Feedback:

> - Feature: being able to specify a generic `gpu` recipe property (ideally alias by alias), and then defining what this implies in a site specific manner, e.g. via `settings.yaml` (e.g. `--nv` or `--rocm` depending on the available hardware)

This is a great idea - if we can define generic recipe properties, we can ask to build a recipe with them. An easy way to do this would be akin to how spack adds variants, e.g.,

```bash
$ shpc install biocontainers/fastqc+gpu
```
Would say "use the settings.yaml file to look for the "gpu" variant, and add these properties to the build. The question comes down to how to add them to the template - right now you can add options to a recipe (e.g. extra flags like --nv) but probably we would want to support different named property types, like:

```
# settings.yaml
properties:
  options:
   gpu: --nv
```
Would this be something that needs more logic than that? E.g., how to distinguish nv or rocm? And would the container be installed with this hard coded, or would the idea be to try and support runtime differences? (I'm not sure how I would do the latter because the module template is generated once. I can maybe look into if / else statements for templates, is that an idea?

marcodelapierre:
Ideally I would love to add those extra bits of logic.
Also, let me clarify the context of my thinking: my key idea was to abstract away the need for knowing the gpu specific requirements of the various container runtimes (singularity has --nv and --rocm, other engines might have other requirements...). So I'd like to enable sys-admins and recipe developers to reason in high-level terms:
- sys-admin: what's the gpu configuration in my cluster?
- recipe developer: does my container need a gpu?

Let me share what I had in mind ideally.

In settings.yaml, a `gpu` property, that can have 3 or 4 values:
```
null or ''  # no gpu
nvidia
amd
$variable
```
This way we use settings to define properties of the cluster we're running on.
The only case where we need $variable is a cluster that has both nvidia and rocm partitions, or systems, then one would define the right variable value at runtime (e.g. in the script: variable=nvidia, or variable=amd).

Then, as a 1st round I would think that needing a GPU or not is a property of the container, so this property would go in the container.yaml. I think of 2 possibilities of implementing this:
container.yaml option 1, a simple container property: (defaults to false, or null)
```
gpu: true
```
or option 2, distinguishing between which binaries need gpus and which don't:
```
aliases:
  /x
  /y
gpu-aliases:
  /w
  /ww
```
For instance, the NGC Gromacs container (if I remember correctly) has only some commands that need the GPU.
[...]

PR #367, opened 2021-05-29
Adding support for custom environment file

### LINKED ISSUE #358: Allow for custom environments for containers
Feedback:

> Feature: being able to define environment variables in recipes (example of use case: python container)

This is a fabulous idea! We can do this easily by binding an environment file at runtime. I'm thinking that perhaps there can be different levels of these files - one specific to the container, and one specific to the user.

### LINKED ISSUE #360: List view organization
Feedback:

> - Comment: the output of `list` is currently organised by tool, and then for each of these a comma-list of versions is given; I was wondering about an alternative output format that is easier to copy-paste or parse for use, e.g. listing one tool/version identifier per line

Yes this is a great idea! Could you provide an example here to get me started?

### LINKED ISSUE #361: Install recipe naming: / vs :
Install Recipe naming

> - Comment: I was a bit confused by the fact that all commands except for `install` require to use '/' and not ':' for the tag; I see where it comes from, and right now I can't think of any better alternative though

I will need to think about this and test some things. I also thought this was not ideal but haven't fixed it yet.

vsoch:
This will address the following issues.

### Environment variables

#358 to add support for defining custom environment variables in the container. This means that any container.yaml can define an env section:

```
env:
   maintainer: vsoch
```
And then this is written to the self.settings.environment_file, which defaults to 99-shpc.sh (so it is sourced close to last). This file is automatically added to every install or "add" folder regardless of these variables or not, so an admin can easily add an environment variable to it. The admin can also update the container.yaml and "re-install" (if the container version has not changed  it won't be re-pulled). Then the environment file is bound to the container where it should be in order to be sourced, at /.singularity.d/env/99-shpc.sh.

### Add Help

The add help was not very good - the "paths" argument was duplicated because of using nargs=2. It is now updated to better reflect a sif_path and module_id. This will fix https://github.com/singularityhub/singularity-hpc/issues/369.

### Developer Documentation

I did a small re-organization of the docs - as I was writing new sections I realized that the user docs (e.g., setup, commands) were confounded with more developer docs (writing a container.yaml recipe). So I split them into two files, one for developer docs and one for the same user docs. It's at least a start to make the documentation better!

<a id="container-features"></a>
### Container Features

We now have the ability to define container features! That means that any container.yaml can define a features section. For example, here is how to say "this container supports gpu":

```yaml
features:
  gpu: true
```
And this is done in a general way because we want these recipes to eventually be usable by multiple container technologies (e.g., podman). Then for the settings.yaml file, there is a matching `container_features` section. And this is where the admin of the registry can decide how to set the features. Here is the default:

```yaml
container_features
  gpu: null
```
And then if the admin has nvidia gpus, they would do:
```yaml
container_features
  gpu: nvidia
```
This is also done a bit generally because it's up to the container technology to say "oh, features.gpu defined as nvidia means I should add the `--nv` flag here! This also means that the templates for the module files (originally template.lua and template.tcl) need to be container technology specific. So they are renamed to `singularity.tcl` and `singularity.lua` accordingly. In the future we could have, for example, `podman.lua`  and `podman.tcl`.

### shpc list

The listing now shows full container names and tags, without spaces or multiple tags per container. This is the default:

```bash
$ shpc list
        biocontainers/samtools:v1.9-4-deb_cv1
                        python:3.9.5-alpine
                        python:3.9.2-slim
                        python:3.9.2-alpine
                      dinosaur:fork
                 vanessa/salad:latest
                         salad:latest
      ghcr.io/autamus/prodigal:latest
      ghcr.io/autamus/samtools:latest
        ghcr.io/autamus/clingo:5.5.0
```

And to get back to the original shorter list with multiple tags per line, the user can add `--short`

```bash
$ shpc list --short
        biocontainers/samtools: v1.9-4-deb_cv1
                        python: 3.9.2-slim, 3.9.5-alpine, 3.9.2-alpine
                      dinosaur: fork
                 vanessa/salad: latest
                         salad: latest
      ghcr.io/autamus/prodigal: latest
      ghcr.io/autamus/samtools: latest
        ghcr.io/autamus/clingo: 5.5.0
```
This will close #360.

## Using Tag Notation

The user can now reference an installed container like

```
shpc check container:tag
```
instead of being required to use a /
```
shpc check container/tag
```
This will close #361 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The versions coincide with releases on pip. Only major versions will be released
  - adding namespaces to make module install, show, inspect easier (0.0.24)
  - added support for env section to render to bound environment file
  - added support for container features like gpu
+ - allowing for a `container:tag` convention to be used for commands.
  - list defaults to showing modules/tags one per line, unless --short used
  - container url does not render in docs (0.0.23)
  - addition of tcl modules, removal of un-needed database (0.0.22)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The versions coincide with releases on pip. Only major versions will be released
  - adding namespaces to make module install, show, inspect easier (0.0.24)
  - added support for env section to render to bound environment file
  - added support for container features like gpu
+ - list defaults to showing modules/tags one per line, unless --short used
  - container url does not render in docs (0.0.23)
  - addition of tcl modules, removal of un-needed database (0.0.22)
  - first update of all containers, and bugfix to pull (0.0.21)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The versions coincide with releases on pip. Only major versions will be released
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/master) (0.0.x)
  - adding namespaces to make module install, show, inspect easier (0.0.24)
+ - added support for env section to render to bound environment file
  - container url does not render in docs (0.0.23)
  - addition of tcl modules, removal of un-needed database (0.0.22)
  - first update of all containers, and bugfix to pull (0.0.21)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The versions coincide with releases on pip. Only major versions will be released
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/master) (0.0.x)
  - adding namespaces to make module install, show, inspect easier (0.0.24)
  - added support for env section to render to bound environment file
+ - added support for container features like gpu
  - container url does not render in docs (0.0.23)
  - addition of tcl modules, removal of un-needed database (0.0.22)
  - first update of all containers, and bugfix to pull (0.0.21)
--- a/docs/getting_started/developer-guide.rst
+++ b/docs/getting_started/developer-guide.rst
@@ -1,0 +1,397 @@
+.. _getting_started-developer-guide:
+
+===============
+Developer Guide
+===============
+
+This developer guide includes more complex interactions like contributing
+registry entries and building containers. If you haven't read :ref:`getting_started-installation`
+you should do that first.
+
+
+Creating a Registry
+===================
+
+A registry consists of a database of local containers files, which are added
+to the module system as executables for your user base. This typically means that you are a
+linux administrator of your cluster, and shpc should be installed for you to use
+(but your users will not be interacting with it).
+
+The Registry Folder
+-------------------
+
+Although you likely will add custom containers, it's very likely that you
+want to provide a set of core containers that are fairly standard, like Python
+and other scientific packages. For this reason, Singularity Registry HPC
+comes with a registry folder, or a folder with different containers and versions
+that you can easily install. For example, here is a recipe for a Python 3.9.2 container
+that would be installed to your modules as we showed above:
+
+.. code-block:: yaml
+
+    docker: python
+    latest:
+      3.9.2: sha256:7d241b7a6c97ffc47c72664165de7c5892c99930fb59b362dd7d0c441addc5ed
+    tags:
+      3.9.2: sha256:7d241b7a6c97ffc47c72664165de7c5892c99930fb59b362dd7d0c441addc5ed
+      3.9.2-alpine: sha256:23e717dcd01e31caa4a8c6a6f2d5a222210f63085d87a903e024dd92cb9312fd
+    filter:
+    - 3.9.*
+    maintainer: '@vsoch'
+    url: https://hub.docker.com/_/python
+    aliases:
+      python: python
+
+And then you would install the module file and container as follows:
+
+.. code-block:: console
+
+    $ shpc install python:3.9.2
+
+But since latest is already 3.9.2, you could leave out the tag.
+
+.. code-block:: console
+
+    $ shpc install python/3.9.2
+
+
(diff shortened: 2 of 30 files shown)


marcodelapierre:
General question:

should the variables be defined in the container (using your proposed mechanism), or instead in the modulefile?
I will also reflect on these two possibilities and their implications, and get back here.

vsoch:
@marcodelapierre my instinct is that it's easier to define them in the module file, the reason being that the maintainer of the install can very easily open up the file and add them without needing to know how to write modulefiles. It also would support doing things like sourcing a file or printing, which wouldn't be possible to do via just an environment variable defined in the module file. Using the bound file also means that we have control over the order of their export. The name I chose should be sourced last, but it could be customized (it's the "environment_file" in settings) to be sooner. environment variables defined for the container at runtime are always (I think) last except for a couple of special cases (we would have to test it). I suspect that if a user of the container module would want to add more environment variables, they would need to export `SINGULARITYENV_X` variables, and (we would again need to test) they could possibly override the ones in the env file.

marcodelapierre:
One thing, as mentioned in #359,
I would use `system_features` (or similar) in settings.yaml, and `features` in containers.yaml, to reduce typing requirements when writing recipes...?

marcodelapierre:
On the strategy for environment variables, let me clarify my thoughts as my previous message was really unclear.
I am thinking about 2 ways of defining the variables through the modulefile:

1. what you've done in the current implementation, i.e. writing a sourceable script that is then bind mounted in the module aliases
2. defining the variables as SINGULARITYENV_var=value in the module

I get your point that a bash script is friendlier to inspect; a module file can be inspected with `module show`, but the syntax would be Lua/Tcl, not bash; also SHPC variables contain a lot of information, making inspection harder.

One comment: maybe define something like `shpc get --environment_file` / `shpc get -e` to print the file location (same as plain `shpc get` for images), or cat it straight away.

The only thing I'd like to further inspect is the order of export; from singularity 3.6.x on, when using 99-shpc.sh, the definitions are not overridable by user via --env or SINGULARITYENV.
Ideally, it would be good if users where able to override those if they want to (although not strictly required).

I was having a look at the environment precedence rule at https://singularity.hpcng.org/user-docs/3.7/environment_and_metadata.html#environment-variable-precedence, and did some tests, but it's not obvious to me yet how such override can be achieved.   

Thoughts?

marcodelapierre:
Now testing this PR:

1. `shpc install <..>`
(This is true also for previous versions)
Output has `Module biocontainers-fastqc/v0.11.5 was created` 
Shouldn't the hash `-` be a slash `/` instead?

2. `shpc list` (plus `--short`): very useful to me

3. `shpc add -h`: help output now clearer

4. Testing the `name:version` syntax with various sub-commands: 
  * working: shell, inspect, get, check, 
  * not working : uninstall


Suggested actions: 
* Fix typo for 1. 
* Fix colon syntax for 4., `shpc uninstall`

vsoch:
> I would use system_features (or similar) in settings.yaml, and features in containers.yaml, to reduce typing requirements when writing recipes...?

This should already be done (see this [updated recipe](https://github.com/singularityhub/singularity-hpc/blob/6ad00314c62c74ad6d1a8e07dc84c7581744abb8/registry/nvcr.io/hpc/gromacs/container.yaml)), but container_features instead of system_features. The thing is, any feature that you can think of for a container (even if a system feature) still can be represented in that group. Container features is an object that can take as many keys as we like to support different kinds of features.

> defining the variables as SINGULARITYENV_var=value in the module

The reason I think we shouldn't do this is because it's the only means for a user (who cannot change or access the installed module) to define a variable at runtime.

> a. One comment: maybe define something like shpc get --environment_file / shpc get -e to print the file location (same as plain shpc get for images), or cat it straight away.

Ok this will be added.

> b. Another comment: when envs are defined, I would probably have shpc add an additional line in the modulefile, help section (both lua and tcl); another way for users to know the location of the environment script.

The envfile is shown as a bind, is that sufficient? We could also remove the bind and just state "This is the envionrment file used across run, exec, shell"

> c. I tested the env feature, and noticed that the sourceable script has a /bin/sh shebang, whereas in the modulefiles you are making sure that /bin/bash is always used; I know they should be equivalent in most practical cases, but maybe it's still worth making the script consistent.

I was following singularity convention here (always uses shell) but I can update to honor the user's request for shell. Note that the module files are already honoring this setting. I'll add a test_shell because I can see wanting it to be different.

> d. The last thing I'd like to further inspect is the order of export; from singularity 3.6.x on, when using 99-shpc.sh, the definitions are not overridable by user via --env or SINGULARITYENV. I think it would be good if users where able to override those if they want to (what do you think?).

I agree - I am not actually sure how singularity currently does this (maybe we can post an issue / discuss in the slack?) We definitely want this! Are you a part of the singularityce slack?

Note that these changes will be pushed to the add/podman branch!

marcodelapierre:
> Suggested actions:
> 
> Fix typo for 1.
> Fix colon syntax for 4., shpc uninstall and shpc test

In addition to these fixes above, there's one more that is required.
`shpc uninstall` only deletes containers, not modules (in this scenario I am using distinct locations):

```
$ shpc uninstall biocontainers/fastqc/v0.11.5
Are you sure you want to uninstall $container_base/biocontainers/fastqc/v0.11.5, and all content below it? y
$container_base/biocontainers/fastqc/v0.11.5 and all subdirectories been removed.

$ ls modules/biocontainers/fastqc/
v0.11.5
```

vsoch:
okay this was a bug - now fixed.

vsoch:
These commits should be in main now, closing here!

PR #384, opened 2021-06-05
adding support for Docker as a new container technology

vsoch:
I'm not sure if this will work out of the box, but docker is fairly similar to podman so it shouldn't be too hard to get working! Since Docker came first, it will provide the "base" that both it and podman use.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         module: ["lmod", "tcl"]
-        container_tech: ["podman", "singularity"]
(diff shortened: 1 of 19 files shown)


vsoch:
Update - podman seems to allow `-it` even if it's not a terminal (warning) but docker does not (error) so we need a configuration setting to disable it for the test. I just added one, should test next.

marcodelapierre:
A nice one with Docker is to always enable `-i`, which allows piping in/out the container.
In the end, `-t` is probably only needed for the <container>-shell alias

vsoch:
This is likely imperfect, but since there is a bug in 0.0.25 and this might work better, going to merge and release.

PR #388, opened 2021-06-06
Improve tcl/lua module code & test

xdelaruelle:
Hi singularity-hpc team! I have looked at the module templates and found that I could help a little bit.

This pull request includes several commits to fix or improve either the Lua or Tcl modulefile generation. I have improved the module tests to check if things were correctly set and fix the issues found, especially for the code generated for Csh-kind shells.

My initial goal was to update the tcl module template to use shell function on Bash shell in Tcl modules, but I found [an issue](https://github.com/cea-hpc/modules/issues/401) with the shell function definition in Environment Modules when the function name includes a `-` character. So for Environment Modules <4.8 it is safe to stay with shell alias.

Hope you will find this PR useful.

### DIFF
--- a/shpc/main/modules/templates/singularity.lua
+++ b/shpc/main/modules/templates/singularity.lua
@@ -46,13 +46,12 @@ local MODULEPATH="{{ module_dir }}"
 setenv("SINGULARITY_SHELL", "{{ singularity_shell }}")
 setenv ("SINGULARITY_OPTS", "")
 setenv ("SINGULARITY_COMMAND_OPTS", "")
-setenv ("SINGULARITY_COMMAND_ARGS", "")
 
 -- interactive shell to any container, plus exec for aliases
 local containerPath = '{{ container_sif }}'
 local shellCmd = "singularity ${SINGULARITY_OPTS} shell ${SINGULARITY_COMMAND_OPTS} -s {{ singularity_shell }} {% if features.gpu %}{{ features.gpu }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }}{% endif %} " .. containerPath
 local execCmd = "singularity ${SINGULARITY_OPTS} exec ${SINGULARITY_COMMAND_OPTS} {% if features.gpu %}{{ features.gpu }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }}{% endif %} "
-local runCmd = "singularity ${SINGULARITY_OPTS} run ${SINGULARITY_COMMAND_OPTS} {% if features.gpu %}{{ features.gpu }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }}{% endif %} " .. containerPath .. " ${SINGULARITY_COMMAND_ARGS}"
+local runCmd = "singularity ${SINGULARITY_OPTS} run ${SINGULARITY_COMMAND_OPTS} {% if features.gpu %}{{ features.gpu }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }}{% endif %} " .. containerPath
 local inspectCmd = "singularity ${SINGULARITY_OPTS} inspect ${SINGULARITY_COMMAND_OPTS} " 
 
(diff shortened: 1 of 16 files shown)


shpc/main/modules/templates/singularity.lua
```diff
@@ -46,6 +46,7 @@ local MODULEPATH="{{ module_dir }}"
 setenv("SINGULARITY_SHELL", "{{ singularity_shell }}")
 setenv ("SINGULARITY_OPTS", "")
 setenv ("SINGULARITY_COMMAND_OPTS", "")
+setenv ("SINGULARITY_COMMAND_ARGS", "")
```
vsoch, line 49: Ah this is a good catch, and I think it's a typo on my part to have `SINGULARITY_COMMAND_ARGS`
```suggestion
```

shpc/main/modules/templates/singularity.lua
```diff
@@ -72,10 +73,10 @@ if (myShellName() == "bash") then
 end{% endif %}
 
 -- A customizable exec function
-set_shell_function("{|module_name|}-exec", execCmd .. containerPath .. " ${SINGULARITY_COMMAND_ARGS}  $@",  execCmd .. containerPath .. " ${SINGULARITY_COMMAND_ARGS} $*")
+set_shell_function("{|module_name|}-exec", execCmd .. containerPath .. " ${SINGULARITY_COMMAND_ARGS}  $@",  execCmd .. containerPath .. " ${SINGULARITY_COMMAND_ARGS}")
```
vsoch, line 76:
```suggestion
set_shell_function("{|module_name|}-exec", execCmd .. containerPath .. " ${SINGULARITY_COMMAND_ARGS}  $@",  execCmd .. containerPath)
```

shpc/main/modules/templates/singularity.tcl
```diff
@@ -40,8 +40,9 @@ proc ModulesHelp { } {
 }
 
 # Environment
-set ::env(SINGULARITY_OPTS) ""
-set ::env(SINGULARITY_COMMAND_OPTS) ""
+setenv SINGULARITY_OPTS ""
+setenv SINGULARITY_COMMAND_OPTS ""
+setenv SINGULARITY_COMMAND_ARGS ""
```
vsoch, line 45:
```suggestion
```

shpc/main/modules/templates/singularity.tcl
```diff
@@ -72,23 +73,23 @@ set runCmd "singularity \${SINGULARITY_OPTS} run \${SINGULARITY_COMMAND_OPTS} {%
...
 
 # exec functions to provide "alias" to module commands
 {% if aliases %}{% for alias in aliases %}
-set-alias "{{ alias.name }}" "${execCmd} {% if alias.options %} {{ alias.options }} {% endif %} ${containerPath} {{ alias.command }} $*"
+set-alias {{ alias.name }} "${execCmd} {% if alias.options %} {{ alias.options }} {% endif %} ${containerPath} {{ alias.command }}"
 {% endfor %}{% endif %}
 
 # A customizable exec function
-set-alias "{|module_name|}-exec" "${execCmd} ${containerPath} \${SINGULARITY_COMMAND_ARGS} $*"
+set-alias {|module_name|}-exec "${execCmd} ${containerPath} \${SINGULARITY_COMMAND_ARGS}"
```
vsoch, line 85: Here as well, if the user wants to give args to the alias they can just be provided! (This was also my mistake, sorry about that!)

```suggestion
set-alias {|module_name|}-exec "${execCmd} ${containerPath}"
```

vsoch (commented):
I am so grateful to get your expertise!  The tests look good, and I just tested quickly on a cluster and the issues I'd be concerned about aliases not getting commands) are not of concern. I did a few edits to remove a mistake on my part for adding a `SINGULARITY_COMMAND_ARGS` that isn't needed, and other than that this should be good to merge! There is one bug I found with the docker executor just now (it requires it for podman which we don't want to do) so after merge here I'll fix that and then draft a new release.

And I will capitalize Lmod and Environment Modules correctly from now on :laughing:

shpc/main/modules/templates/singularity.lua
vsoch: I'm thinking not - I removed it since you originally removed `$*` (but let me know if this is wrong and I should add it back).

xdelaruelle: The `$@` is needed in shell function as functions are not expanded like aliases are.

The `set_shell_function` of Lmod use its second argument to define a shell function if current shell is an sh shell, third argument is the code to set a shell alias if current shell is a csh shell.

vsoch: Gotcha, added it back. And the csh doesn't need to accept args?

xdelaruelle: For csh an alias is set and as any alias it is expanded in place so the remaining word after this alias name are de-facto its arguments

vsoch:
Everything looks great - let's merge in these improvements! Thank you again for your expertise here - the files look much better.

xdelaruelle:
Happy to help

Issue #378, opened 2021-06-03
Rename namespace to repository 


marcodelapierre:
Just capturing (read: lazily copy-pasting) our Slack chat here.

V: for the singularity versions I added the module version to the executable name, and as you pointed out i didn't to podman's. But I'm rethinking this, and I'm wondering if in fact we shouldn't include the version? E.g., would there ever be a use case of having more than  one python (or something named the same thing) defined twice? It would be a shorter/cleaner command if we remove versions, and I can add more conflicts() clauses.

V: oh one more idea for the command naming! We could have a naming scheme that defaults to `{name}-{version}` (like with spack modules) so the person deploying the modules can decide! Maybe default would just be name though.

marcodelapierre:
Marco: I like the idea of the naming scheme! And I think you're right, I also agree with your other points, rather than fixing the version missing in podman, it's probably more user friendly to get rid of the version in singularity. So I'd suggest `{name}`  as the default prefix.

marcodelapierre:
Also, talking of naming conventions, I was thinking whether it'd be possible to allow a flexible naming scheme for modules, too.

E.g., say we install `quay.io/biocontainers/samtools/1.9` , I'd love being able to ditch the first part of the repo and just have `samtools/1.9` . 
I see a risk in clashing names between different recipes, so I'd probably leave the current behaviour as default. 
But for clusters that know they're not going to mix things, it'd be handy to have the option.

vsoch:
@marcodelapierre I was thinking of this too, and the reason I haven't done it yet is because we'd need another way to identify the module once it's installed. Right now I'm just parsing the directory structure as an identifier, and I figured that since LMOD has autocompletion it wasn't a big deal to autocomplete. But I did think it would be cool to be able to do:

```bash
$ shpc install quay.io/biocontainers/samtools/1.9 samtools/1.9
```
Arguably you could do any path that you want!

```bash
$ shpc install quay.io/biocontainers/samtools/1.9 bio/samtools/1.9
```

It would be up to the admin to make sure there aren't namespace clashes, and actually we could keep the modules installed and registry files separate, but I think the main issue comes with `shpc check`, which maybe isn't a useful command, but we do match modules installed to the install folder by way of the path.

[...]

PR #397, opened 2021-06-07
Updates to several container and module bases

### LINKED ISSUE #391: Minimal comments on Podman/Docker support
Hi!
I have just run some quick tests with Podman on Ubuntu 18, and they have worked well!

I just have a couple of minimal comments on the modulefile layout (had a look at LUA modules);
I have used my usual `biocontainers/samtools:v1.9-4-deb_cv1` as a test.

- Some typos in `help screen` -- apply also for Docker

    - `-v ${PWD} -w ${PWD}` is defined also for `<tool>-shell` and `<tool>-exec` but not documented

    - `--entrypoint /bin/sh` is defined for `<tool>-shell` but not documented

    - for `<tool>-exec`, help says `--entrypoint /bin/sh`, but instead it should be `--entrypoint ""`

    - for aliases, e.g. :
      ```
       - samtools:
             podman run --rm -it --entrypoint /usr/bin/samtools --env-file /home/ubuntu/shpc-v0027/modules/biocontainers/samtools/v1.9-4-deb_cv1/99-shpc.sh  -v ${PWD} -w ${PWD} <container> ""
      ```
      Instead of `<container> ""`, the very last bit of this snippet should be `<container> "$@"`.

Only for Podman:

- I notice that the variables here are called `DOCKER_OPTS` and `DOCKER_COMMAND_OPTS`; shouldn't they contain `PODMAN` instead?

PS: the final implementation you used for `entrypoint` is the cleanest possible, love it!

### LINKED ISSUE #392: shpc test fails with podman and docker
(with and without version tag)

seems like it's looking for the SIF file

```
$ shpc test biocontainers/samtools
v1.9-4-deb_cv1: Pulling from biocontainers/samtools
Digest: sha256:da61624fda230e94867c9429ca1112e1e77c24e500b52dfc84eaf2f5820b4a2a
Status: Image is up to date for biocontainers/samtools:v1.9-4-deb_cv1
docker.io/biocontainers/samtools:v1.9-4-deb_cv1
Module biocontainers/samtools:v1.9-4-deb_cv1 was created.
  File "/data/work/shpc/shpc-v0027/bin/shpc", line 33, in <module>
    sys.exit(load_entry_point('singularity-hpc', 'console_scripts', 'shpc')())
  File "/data/work/shpc/shpc-v0027/shpc/client/__init__.py", line 349, in run_shpc
  File "/data/work/shpc/shpc-v0027/shpc/client/test.py", line 23, in main
    template=args.template,
  File "/data/work/shpc/shpc-v0027/shpc/main/client.py", line 148, in test
    assert os.path.exists(sif)
AssertionError

```

### LINKED ISSUE #393: shpc uninstall fails with docker/podman
```
$ shpc uninstall biocontainers/samtools:v1.9-4-deb_cv1
Multiple tags found for biocontainers/samtools/v1.9-4-deb_cv1: 99-shpc.sh, module.lua.
```

### LINKED ISSUE #394: Enforce host user when using Docker as runtime
Hi Vanessa,

based on my experience, I'd recommend adding the following flags to all aliases and commands when using Docker:

```
-u $(id -u):$(id -g)
```

This should work with most Linux'es (tested with Ubuntu, Centos, SLES, CLE).
In this way, files created in the host by the container will have the right ownership.

### LINKED ISSUE #395: Question on conflict modules in generated modulefiles
I was testing a modified `biocontainers/samtools:v1.9-4-deb_cv1` with a 2nd alias, just to check how aliases looked like (I added `bamtools: /usr/bin/bamtools`).

The aliases looked right both in Lua and Tcl, however I noticed that the conflict list grew with that additional alias, both for Lua and Tcl; for Lua:
```
conflict(myModuleName(),"samtools","bamtools")
```
for Tcl:
```
conflict biocontainers/samtools
conflict samtools
conflict bamtools
```

What is the rationale behind adding a conflict for each alias name? 
Also, why does the list look different for the two languages?

Thanks,

### LINKED ISSUE #396: Notation for Lua/Tcl module languages
These are just minor notation/documentation points:

1. in `settings.yaml`, there is the following:
```
# set a default module system
# Currently only lmod is supported. To request an additional system, 
module_sys: lmod
```
This should probably be updated to include tcl

2. I am wondering whether `lua` should be allowed as an equivalent option to `lmod`; the former could be more natural to use to some people, when having to pick which *language* they want SHPC to use for the modulefiles

vsoch:
This PR include several bug fixes or updates, specially:

 - Add test for docker and podman to fix #392 
 - various module template updates to close #391
 - namespace as format string for command named renamed to repository to address #378
 - updates to comment in settings to close #396
 - shpc test/uninstall should be run for all tests
 - bug with uninstall fixes #393 
 - adding user and group ids to docker and podman commands to close #394
 - remove conflict with entire module name, should only be for aliases to close to close #395


marcodelapierre:
On module name conflict, as I mentioned in #395, It's probably good to leave the full name, as it prevents from accidentally loading two versions of the same SHPC container module. So ... can it be added for Lua, too?

On the Docker/Podman optional env variables (as in my comment to #391), I was wondering ... instead of using DOCKER for Podman, too ... how about using PODMAN for Docker ...? too daring ...? after all SHPC is more focused on HPC, Docker is kind of a free add-on feature

marcodelapierre:
`{{ namespace }}` should be changed to `{{ repository }}` also in the comment in `settings.yaml`

vsoch:
> On module name conflict, as I mentioned in #395, It's probably good to leave the full name, as it prevents from accidentally loading two versions of the same SHPC container module. So ... can it be added for Lua, too?

Added back! For Lua I didn't add `{{ name }}` because I thought that `myModuleName()` would translate to that. But if it does not (?) I definitely can.

> On the Docker/Podman optional env variables (as in my comment to #391), I was wondering ... instead of using DOCKER for Podman, too ... how about using PODMAN for Docker ...? too daring ...? after all SHPC is more focused on HPC, Docker is kind of a free add-on feature

Let's be daring!

marcodelapierre:
Testing related GH issues, and new features. 

All good but this one:

**Help block** in Podman modulefile:
* `-v ${PWD} -w ${PWD}` missing in `<tool>-run` and `<tool>-shell`
* `--entrypoint /bin/sh` missing for `<tool>-shell` (I guess that `/bin/sh` is actually a variable taken from settings.yaml)

marcodelapierre:
Note that issues #378 and #398 are also solved by this PR :-)

vsoch:
@marcodelapierre I don't see those issues in either docker.tcl or docker.lua - could you comment directly on the code here to point them out?

marcodelapierre:
@vsoch it's only for Lua, in lines from 18 on of `docker.lua`:
```
 - {|module_name|}-run:
       {{ command }} run -i{% if tty %}t{% endif %} -u `id -u`:`id -g` --rm {% if envfile %}--env-file {{ module_dir }}/{{ envfile }}{% endif %} {% if bindpaths %}-v {{ bindpaths }} -v ${PWD} -w ${PWD} {% endif %}<container> "$@"
 - {|module_name|}-shell:
       {{ command }} run -i{% if tty %}t{% endif %} -u `id -u`:`id -g` --rm {% if envfile %}--env-file {{ module_dir }}/{{ envfile }}{% endif %} {% if bindpaths %}-v {{ bindpaths }} --entrypoint {{ shell }} -v ${PWD} -w ${PWD} {% endif %}<container>
 - {|module_name|}-exec:
       {{ command }} run -i{% if tty %}t{% endif %} -u `id -u`:`id -g` --rm --entrypoint "" {% if envfile %}--env-file {{ module_dir }}/{{ envfile }}{% endif %} {% if bindpaths %}-v {{ bindpaths }} {% endif %} -v ${PWD} -w ${PWD} <container> "$@"
```

For shell and exec,
There's a misplaced ` {% endif %}` that should go straight after `{% if bindpaths %}-v {{ bindpaths }}`

marcodelapierre:
In the TCL module files for Podman, I actually notice other issues (had not checked before):

help screen:
- missing `--rm` for alias commands
- for `<tool-exec>` entrypoint should NOT be `--entrypoint /bin/sh` but instead `--entrypoint ""`

actual command definitions:
- `set execCmd` misses `-v $workdir -w $workdir`
- alias commands should have NOTHING after `${containerPath}` (right now the command is repeated, but it's already in the entrypoint)

vsoch:
> alias commands should have NOTHING after ${containerPath} (right now the command is repeated, but it's already in the entrypoint)

This is because some aliases have other arguments following the executable. If it's repeated it's a bug with deriving alias.entrypoing vs. alias.args, let me take a look.

PR #409, opened 2021-06-22
added quay.io/biocontainers tools

audreystott:


vsoch:
These are beautiful @audreystott ! It looks like my container module test isn't set up for cloning from other repos (your PR is from a branch called main so it is hiccuping) but I've looked over the recipes and they look good, so I think it's okay to merge. We can always adjust anything that needs to be tweaked as you are using the containers.

Thank you for adding these! I did notice that biocontainers seemed to jump to quay.io but didn't add a lot of their containers.

registry/quay.io/biocontainers/spades/container.yaml
```diff
@@ -0,0 +1,11 @@
+docker: quay.io/biocontainers/spades
+latest:
+ "3.15.2": "sha256:250657af71d37fafdff81af644e92bf84485d4d9b6894f58ff84efdc1a6ec9eb"
+tags:
+ "3.15.2": "sha256:250657af71d37fafdff81af644e92bf84485d4d9b6894f58ff84efdc1a6ec9eb"
+ "3.15.0": "sha256:4cc7e41f53d3effb25b59afedc3fbd2d3e08fff449a507ecfa6731ff6d81ff79"
+maintainer: "@audreystott"
+description: "SPAdes – St. Petersburg genome assembler – is an assembly toolkit containing various assembly pipelines.."
```
vsoch, line 8:
```suggestion
description: "SPAdes – St. Petersburg genome assembler – is an assembly toolkit containing various assembly pipelines."
```

audreystott:
Thanks Vanessa! Should I rename my branch name to something else? Yeah there are quite a number of them under quay.io, and we tend to use these ones at our centre - so this works as a double-edged sword for us to add it to your repo.

vsoch (approved):

vsoch:
@audreystott if you open another PR soon it would be good to do a branch that isn't main, and that way we can test a simple install and module load for each one. For git sanity this is a good general practice because main should always match the main of the upstream. For example, I just squashed and merged so the main branch here (upstream to your repo) branch will have a commit that your clone does not!) and if you keep creating new branches off of your main that extra commit will show up as a weird extra commit in future PRs. But for the longer run I will fix the label test so that it can run from another fork's main. Sorry for that trouble - you're the first to do exactly that!

And your newly added containers are already showing up in the library - https://singularityhub.github.io/singularity-hpc/ woohoo! :)

PR #424, opened 2021-07-20
Escape `$` for TCL modules (don't deref in TCL)

zbeekman:
Example to fix (at least for TCL modules) for jupyter/minimal-notebook. Untested with LMOD.

TCL modules will dereference (expand) things that look like bash
variables and "new-style" command substitutions.


zbeekman:
This is obsoleted by #425

PR #425, opened 2021-07-20
Fixing bug that envars need to be escaped for tcl

vsoch:
Also updating recipes for jupyter containers to create a temporary directory to bind to /var/run.

@zbeekman I think this should fix both the errors you were running into!


zbeekman:
@vsoch I can confirm that this resolved the TCL modules problem and I was able to run the jupyter/minimal-notebook `run-notebook` alias from the JOSS paper example.

PR #427, opened 2021-07-20
changes in bib to try to fix current and future issues

danielskatz:


vsoch (approved):
Thanks @danielskatz ! I've never had issues with abstracts and JoSS but there is a first time for everything!

danielskatz:
I don't think the abstracts are the problem - my guess is the ~ in the month was.  But the abstracts will be added to the XML submitted to crossref for entries without DOIs, which will be messy for crossref to deal with, so I thought we could simplify that too at the same time

PR #437, opened 2021-08-12
Docs/feedback marco august 2021

marcodelapierre:
Hi @vsoch ,

I did a read through of the documentation, as I remembered some small bits that could be cleaned.
It's small work, but I hope it can contribute to the cause!

Also, note there's an empty code block in `docs/getting_started/user-guide.rst`, around line 965 (just before the Add command section). I could not figure out the missing bit myself.


vsoch (approved):
Thank you!

PR #443, opened 2021-09-03
Inspect other bug fixes - podman_options rather than docker_

marcodelapierre:


vsoch:
Hmm I think I'd tend to agree that we are targeting HPC, but we can't be certain yet that Podman is really going to be "taking off." I think most people are more familiar with Docker, and then those that are familiar with podman know the two are interchange-able. 

I would just move to the other PR, but I realize that we would have a contradictory design - Docker is considered the "base" https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/container/docker.py#L2 and podman is on top of that. By having `podman_args` we would imply the opposite I think.

marcodelapierre:
Agree! closing

Issue #454, opened 2021-11-18
Passing commandline arguments to singularity containers?


jacquikeane:
Not sure if this is a bug or something I am doing incorrectly. After using shpc to install bwa and running bwa index it returns me the output as if I had just run bwa.

![Screenshot from 2021-11-18 16-49-10](https://user-images.githubusercontent.com/872885/142459665-14964739-1221-40bd-9b6f-de6fa7f20797.png)

Below is a snippet from the module.tcl generated by shpc for the bioinformatics software bwa but this will apply to all software applications. How are the commandline arguments passed to the singularity container? Should the function and alias commands include ''"$@""? Or is this handled somewhere else? 

![Screenshot from 2021-11-18 16-44-19](https://user-images.githubusercontent.com/872885/142458890-c6974ba9-1e47-4b59-bf93-6e313ad96dec.png)

vsoch:
This looks like a bug - @marcodelapierre added these new functions rather recently, and indeed we are missing the argument parsing at the end! https://github.com/singularityhub/singularity-hpc/blob/1b800836e8bd53150eba9da0474e0506b8b3318f/shpc/main/modules/templates/singularity.tcl#L81. We can definitely fix this up asap.

vsoch:
@jacquikeane do you have the same issue for any of the other interactions (e.g., the -exec functionality) or is it just the aliases?

jacquikeane:
Ah excellent thanks!

jacquikeane:
> @jacquikeane do you have the same issue for any of the other interactions (e.g., the -exec functionality) or is it just the aliases?

I have only tested this with bash. And yes I am seeing the issue with all the other interactions.

marcodelapierre:
@vsoch indeed seems just like the wildcard missing from the aliases/functions - I have tested lmod extensively, but clearly tcl not us much 

going to test your PR asap

vsoch:
Thank you! I added `$*` where I thought it would be appropriate, and my questions are:

- did I miss any commands?
- do we need to account for the shell? E.g., in previous versions we had one version for csh and bash.

marcodelapierre:
Will have a look as soon as I can.

Why didn't `$@` work? I am surprised!

vsoch:
[...]

PR #455, opened 2021-11-18
Add/function args

vsoch:
Here is a fix to test #454. I'm also wondering if this is an issue with versioning? Perhaps a newer version would be able to accept these args - indeed Singularity is pretty seamless so I thought it would work as is (and if I remember we used to have these and possible removed them?)

ping @marcodelapierre and @xdelaruelle  who I think removed all of the argument passing in https://github.com/singularityhub/singularity-hpc/pull/388/files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The versions coincide with releases on pip. Only major versions will be released
 
 ## [0.0.x](https://github.scom/singularityhub/singularity-hpc/tree/master) (0.0.x)
  - `.version` file should be written in top level of module folders [#450](https://github.com/singularityhub/singularity-hpc/issues/450) (0.0.35)
+  - tcl module functions need `$@` to handle additional arguments
  - Tcl modules use shell functions for bash, to export to child shells (0.0.34)
    - fixed missing singularity -B flag for custom home feature
    - fixed singularity.tcl to always replace $ with \$ for custom home feature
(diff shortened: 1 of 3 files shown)


marcodelapierre:
going to check ASAP

xdelaruelle:
Hi @vsoch. Hi @marcodelapierre. In the change proposed there is no need for the ``$*`` addition for the shell alias definition.

Shell alias are substituted in place on shells (either sh or csh family), thus the arguments set on the command line will be left untouched. So there is no need to append the list of argument at the end of the shell alias definition.

So the lines starting with ``set-alias`` should be left untouched by this commit.

Of course for the shell function definition there should be a `$*` or `$@` somewhere to pass the list of args. So this part of the change seems correct to me.

vsoch:
@xdelaruelle thank you! But if that is the case, why is the alias not receiving additional input args as reported in #454 ?

marcodelapierre:
maybe it's just a naming typo, they said alias but meant shell functions (default for bash)?
@jacquikeane, in #454 were you using a bash shell with the latest SHPC version?

xdelaruelle:
As far as I understand the user report, she is using bash, so the code snipped screenshot of module.tcl says a `bwa` shell function is defined. And this shell function does not handle arguments.

marcodelapierre:
Many thanks for pointing that out @xdelaruelle !
@vsoch , could you try to go back to `$@` for the shell function cases in those templates ? I think it should work, and is probably more appropriate in terms of behaviour, as well as consistent with the lua/lmod version

vsoch:
okay restored so that it's removed from set-alias. It could just be that she hit the error where the function is defined, and it was missing the argument capture.

marcodelapierre:
tested quickly, seems good to go!

vsoch:
@jacquikeane does this fix your issue?

PR #469, opened 2022-01-18
Fix: parsing of config parameters for set/get

marcodelapierre:
Issue: shpc was failing to set nested configuration parameters:
```
$ shpc config set container_features:gpu:amd
container_features:gpu:amd cannot be added to config: 'gpu:amd' is not of type 'object'
```

This was because the original parsing would result in the following, under the hood:
```
key = "container_features"
value = "gpu:amd"
```

This PR fixes this issue:

```
$ shpc config set container_features:gpu:amd
Updated container_features:gpu to be amd
```

Under the hood:
```
key = "container_features:gpu"
value = "amd"
```

And checking:
```
$ shpc config get container_features
container_features             ordereddict([('gpu', 'amd'), ('x11', None), ('home', None)])
$ shpc config get container_features:gpu
container_features:gpu         amd
```

Important NOTE: I am unsure about the change I made for the `get` method in the second commit. I got rid of a get call using a 2nd `default` argument. Unsure whether this can result in bad behaviours.

### DIFF
--- a/shpc/client/config.py
+++ b/shpc/client/config.py
@@ -41,7 +41,8 @@ def main(args, parser, extra, subparser):
                     % param
                 )
                 continue
-            key, value = param.split(":", 1)
+            value, key = param[::-1].split(":", 1)
+            key, value = key[::-1], value[::-1]
             if command == "set":
                 cli.settings.set(key, value)
                 logger.info("Updated %s to be %s" % (key, value))
--- a/shpc/client/config.py
+++ b/shpc/client/config.py
@@ -1,5 +1,5 @@
 __author__ = "Vanessa Sochat"
-__copyright__ = "Copyright 2021, Vanessa Sochat"
+__copyright__ = "Copyright 2021-2022, Vanessa Sochat"
 __license__ = "MPL 2.0"
 
 import shpc.defaults as defaults
@@ -41,8 +41,8 @@ def main(args, parser, extra, subparser):
                     % param
                 )
                 continue
-            value, key = param[::-1].split(":", 1)
-            key, value = key[::-1], value[::-1]
+
+            key, value = param.split(":", 1)
             if command == "set":
                 cli.settings.set(key, value)
                 logger.info("Updated %s to be %s" % (key, value))
--- a/shpc/main/settings.py
+++ b/shpc/main/settings.py
@@ -183,7 +183,7 @@ class SettingsBase:
             logger.exit("You cannot use 'set' for a list. Use add/remove instead.")
 
         # This is a reference to a dictionary (object) setting
-        if ":" in value:
+        if isinstance(value, str) and ":" in value:
             subkey, value = value.split(":")
             self._settings[key][subkey] = value
         else:
--- a/shpc/main/settings.py
+++ b/shpc/main/settings.py
@@ -1,5 +1,5 @@
 __author__ = "Vanessa Sochat"
-__copyright__ = "Copyright 2021, Vanessa Sochat"
+__copyright__ = "Copyright 2021-2022, Vanessa Sochat"
 __license__ = "MPL 2.0"
 
 
@@ -183,8 +183,8 @@ class SettingsBase:
             logger.exit("You cannot use 'set' for a list. Use add/remove instead.")
 
         # This is a reference to a dictionary (object) setting
-        if ":" in key:
-            key, subkey = key.split(":")
+        if ":" in value:
+            subkey, value = value.split(":")
             self._settings[key][subkey] = value
         else:
             self._settings[key] = value
--- a/shpc/main/settings.py
+++ b/shpc/main/settings.py
@@ -119,7 +119,7 @@ class SettingsBase:
             key, subkey = key.split(":")
             value = self._settings[key][subkey]
         else:
-            value = self._settings[key]
+            value = self._settings.get(key, default)
         value = self._substitutions(value)
         # If we allow environment substitution, do it
         if key in defaults.allowed_envars and value:
--- a/shpc/main/settings.py
+++ b/shpc/main/settings.py
@@ -114,7 +114,12 @@ class SettingsBase:
         """
         Get a settings value, doing appropriate substitution and expansion.
         """
-        value = self._settings.get(key, default)
+        # This is a reference to a dictionary (object) setting
+        if ":" in key:
+            key, subkey = key.split(":")
+            value = self._settings[key][subkey]
+        else:
+            value = self._settings[key]
         value = self._substitutions(value)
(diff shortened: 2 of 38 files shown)


shpc/client/config.py
```diff
@@ -41,7 +41,8 @@ def main(args, parser, extra, subparser):
                     % param
                 )
                 continue
-            key, value = param.split(":", 1)
+            value, key = param[::-1].split(":", 1)
```
vsoch, line 44: Can you walk me through this logic? It doesn't read clearly for the developer reader to get a sense of what is going on. Perhaps add an example to the comment?

marcodelapierre: Sure, let me explain, and maybe we'll find out there's a neater way to code this.
The old split in line 44 will split by a single colon entry, the first one encountered in the string, to produce a key,value pair.
What we need for the fix is to split by the LAST colon entry. 
So I thought, let's revert the string, split by the first colon entry, and then revert back:
```
# split reversed string by first colon entry, which is the last entry in the original string
value, key = param[::-1].split(":", 1)
# reverse back the (key, value) strings
key, value = key[::-1], value[::-1]
```

Does it make sense?
Do you think there's a cleaner way to achieve this?

Thanks for your feedback! :)

shpc/main/settings.py
```diff
@@ -114,7 +114,12 @@ def get(self, key, default=None):
         """
         Get a settings value, doing appropriate substitution and expansion.
         """
-        value = self._settings.get(key, default)
+        # This is a reference to a dictionary (object) setting
+        if ":" in key:
+            key, subkey = key.split(":")
+            value = self._settings[key][subkey]
```
marcodelapierre, line 120: @vsoch , to be honest, I find this edit of mine to be a bit dirty: the case for a `value` of type dictionary uses a different method than the standard, non-dict case. 
I am not sure why the latter uses the `.get()` method, unlike for instance implementation in the `add` function further down in this source file. 

Ultimately, what I wanted to add here was the ability for this `get` function to handle dict type entries in the settings, same as I am achieving for the `set` function with the other edit in this PR.

Do you think there's a better way to add this functionality?

vsoch: ah, so you gave me the example of what doesn't work, so let me inspect that and either make a suggested change or just for comments! I'm actually working late today and need to have dinner (it's 7pm!) but I'll take a look after dinner!

vsoch: haha okay here is what threw me off (confirming now) the first line seems to reverse the whole thing?
```bash
param[::-1]
'dma:upg:serutaef_reniatnoc'
```
Let me see if I can make something work that is a little more straight forward - I'll update the changelog / version / license headers too. back in a bit!

marcodelapierre: yes, the 1st line reverses the whole string, to split with the right colon, and then the 2nd line individually reverses the two substrings

vsoch: okay take a look! So if you look in the settings.py, the logic was there to parse a nested value, but I had the wrong check - checking the key instead of the value (and indeed the nesting would happen in the value `gpu:amd`. So we didn't need to change config.py under client, just that typo. I also took the opportunity to update the license headers because HAPPY NEW YEAR! :partying_face:

shpc/main/settings.py
```diff
@@ -178,8 +183,8 @@ def set(self, key, value):
             logger.exit("You cannot use 'set' for a list. Use add/remove instead.")
 
         # This is a reference to a dictionary (object) setting
-        if ":" in key:
-            key, subkey = key.split(":")
+        if ":" in value:
+            subkey, value = value.split(":")
```
vsoch, line 187: And I could have also put `.split(":", 1)` here, the assumption being that we only have one level of nesting, but I left it like this because I'd rather it error out and give the user a message (and then deal with it) then to silently pass another layer of `:` into a config.

marcodelapierre: Yep, makes sense!

For the `get` function , lines 118-122, does my edit seems sensible to you?

vsoch: everything except for the fact that we want to check the value and not the key - the key (by way of the split on the first :) won't have the second ":" - but I think you probably wrote it given the reverse in config.py, so my changes here should resolve both those.

vsoch: oh no wait I totally missed that - that is wrong too.

shpc/main/settings.py
```diff
@@ -114,7 +114,12 @@ def get(self, key, default=None):
         """
         Get a settings value, doing appropriate substitution and expansion.
         """
-        value = self._settings.get(key, default)
+        # This is a reference to a dictionary (object) setting
+        if ":" in key:
```
vsoch, line 118:
```suggestion
        if isinstance(value, str) and ":" in value:
```

shpc/main/settings.py
```diff
@@ -114,7 +114,12 @@ def get(self, key, default=None):
...
+            key, subkey = key.split(":")
```
vsoch, line 119:
```suggestion
            subkey, value = value.split(":")
```

vsoch (commented):
okay I think this would work? That said we probably should add a test for it. And I'm wondering if when we check for "true" or "false" as a string that should be _after_ this additional value parsing?

shpc/main/settings.py
vsoch: actually let me just edit these locally and add a test - need a few minutes getting ready for bed! I think we might not need to do this, we don't care about the value for get.

shpc/main/settings.py
vsoch: Oh wait I think I understand - you want to do like:

```python
settings.get('container_features:gpu')
# amd
```
vs.

```python
settings.get('container_features')['gpu']
```
I was thinking the second is more straight forward in terms of parsing python, but if you are interested just in the value I can see the first being okay. It's different than the second parsing (set) where the second ":" is in the value but if it works I have no issue with it. Let me know your thoughts!

shpc/main/settings.py
marcodelapierre: I think the test we need is on the lines of `settings.get("container_features:gpu")`

shpc/main/settings.py
marcodelapierre: I think we need whichever implementation makes the following CLI command to work: `shpc config get container_features:gpu`

vsoch: okay then I think what you have is perfecto - checking for the : in the key, and then splitting based on that. I'll add a test for it, and it that passes (and all is okay with you) merge and release.

marcodelapierre:
All looks good to me!

Issue #383, opened 2021-06-04
Allow for script definition instead of alias


marcodelapierre:
**Describe the bug**

Attempting to use command aliases from SHPC modules in child shells (e.g. with Nextflow) causes an error, as they turn out to be undefined there.

First test: bash and LUA

**To Reproduce**

Setup and OK test:
```
shpc install biocontainers/samtools:v1.9-4-deb_cv1

module use <appropriate path>
module load 
biocontainers/samtools/v1.9-4-deb_cv1

type samtools
```

To get the ERROR: launch a new `bash` shell, and try `type samtools` again.

Fix as described here: https://github.com/CHPC-UofU/Lmod-setup/
For each shell function called <FUNCTION>, add:
```
if (myShellName() == "bash") then
 execute{cmd="export -f FUNCTION",modeA={"load"}}
end
```

Second test: bash and TCL

Now, with TCL it's tougher, because it's using `set-alias` ... aliases are worse to make work in child shells than shell functions. 
I tried a few tricks I knew (`bash -i`, `shopt -s expand_aliases`, ..), they did not work.

I see that Env Modules from version 4.2 supports a new feature, `set-function`: `https://modules.readthedocs.io/en/latest/modulefile.html`

However, my old Lmod 7.6.2 does not like it.
[update] : no Lmod version supports it so far: https://lmod.readthedocs.io/en/latest/051_tcl_modulefiles.html

Going ahead to have a look at csh.

marcodelapierre:
Third test:  csh and LUA

seems like csh does not have proper shell functions, so lua falls back to creating aliases here.
Again, could not find a way to have them inherited in sub-processes so far.

marcodelapierre:
It's good that there's a workaround for bash+lua anyway!

marcodelapierre:
Going to document one more possible case, mentioned by the folks above, https://github.com/CHPC-UofU/Lmod-setup/    :  does `mpirun` like shell functions and aliases?

marcodelapierre:
There might be another corner cases, i.e. when the application needs to be run with MPI, e.g. `mpirun` or Slurm `srun`, or whatsoever. 
Looks like these guys don't like shell functions neither (again pointed out first in the link above, https://github.com/CHPC-UofU/Lmod-setup/

Still think it's kind of a corner cases because, at least for what I've seen so far, MPI applications are not deployed in containers that often.

marcodelapierre:
Last thing of this work week...

[...]

PR #473, opened 2022-01-20
New functionality: wrapper scripts

marcodelapierre:
With this PR, I am adding a beta version for the wrapper scripts functionality that I was mentioning in issue #383 . The main scope is to enable interaction of the command aliases with mpirun/srun, which don't work with shell aliases.

In this first iteration, it is an on/off functionality controlled in the settings.yml; I have added a 2nd setting to control the shell type in the wrapper scripts (I just added it in compliance with other shell usages in SHPC):
```
# for container aliases, use wrapper shellscripts instead of shell aliases, defaults to false
wrapper_scripts: false

# shell for wrapper shellscripts
wrapper_shell: /bin/bash
```

Setting `wrapper_scripts` to false (default), gives the traditional behaviour of shpc.
When it is set to true, 2 things change:
1. under the module directory for that container, a bin/ subdirectory is created, containing one wrapper script per command alias
2. modulefile: instead of the command aliases, PATH is prepended with "module_dir/bin"
Note that aliases are still created for the generic commands -exec, -shell and so on.

Support:
- shells: bash, sh, csh
- container techs: singularity, docker/podman
- module systems: lmod, tcl

Implementation notes:
- there are two new templates, `singularity.sh` and `docker.sh`
- edited the util function write_file, to add an option for making the file executable
- most changes are in the container.install() method in `singularity.py` and `docker.py`
- in this first iteration, the code block that generates the wrapper scripts is inside the install() method; it might be turned into a stand-alone method for better modularity and readability
- I have changed the api to install() and add() (singularity only), to pass `wrapper_template` between module and container methods. Maybe this could be made cleaner?

Other notes: 
- are unit tests needed?
- documentation still to be updated

What do you think?


vsoch (commented):
Ah okay I think I get this conceptually! I'm going to need an evening or two to test and review, and order wise I want to get https://github.com/singularityhub/singularity-hpc/pull/472 in first just for review order of operations. So stay tuned! One thing to think about in the meantime is, aside from having the aliases be the driver for the wrapper scripts (e.g., it looks like each alias becomes an executable to basically run a container) what other functionality could be provided? E.g., what would be a different kind of wrapper script, or even one that is custom provided by a user? I will think on this too - so expect to hear from me probably in a few days, latest the weekend!

shpc/main/modules/templates/singularity.tcl
```diff
@@ -74,6 +74,11 @@ set inspectCmd "singularity \${SINGULARITY_OPTS} inspect \${SINGULARITY_COMMAND_
 # set_shell_function takes bashStr and cshStr
 set-alias {|module_name|}-shell "${shellCmd}"
 
+{% if wrapper_scripts %}
+# add path with "alias" wrapper scripts to PATH
+{% if aliases %}prepend-path PATH "{{ wrapper_dir }}"{% endif %}
+
```
vsoch, line 80: I think this is going to create a lot of empty space - if possible let's clean out the extra newlines.

shpc/main/modules/templates/singularity.tcl
```diff
@@ -74,6 +74,11 @@ set inspectCmd "singularity \${SINGULARITY_OPTS} inspect \${SINGULARITY_COMMAND_
 # set_shell_function takes bashStr and cshStr
 set-alias {|module_name|}-shell "${shellCmd}"
```
vsoch, line 75: ah so one interesting thing I just ran into (that I think we should find a workaround for) is for modules to be portable. E.g., right now the module paths are hardcoded, and for this reason I couldn't find the clingo script (I have to run a container to test).

```bash

-- add path with "alias" wrapper scripts to PATH
prepend_path("PATH", "/home/vanessa/Desktop/Code/shpc/modules/ghcr.io/autamus/clingo/5.5.1/bin")
```
What I think we want to do is figure out how to either have a top level config file that sets a variable for the root, OR to calculate on the fly. I haven't found good examples yet (most are for Lua as a script) so i don't have good suggestions. But I do think it would be nicer if we could move an entire module directory without it breaking!

shpc/utils/fileio.py
```diff
@@ -114,12 +115,14 @@ def copyfile(source, destination, force=True):
...
 
 
-def write_file(filename, content, mode="w"):
+def write_file(filename, content, mode="w", exec=False):
     """
     Write content to a filename
     """
     with open(filename, mode) as filey:
         filey.writelines(content)
+    if exec:
```
vsoch, line 124: Let's avoid using subprocess here - maybe try:

```bash
import os
import stat

st = os.stat(filename)
os.chmod(filename, st.st_mode | stat.S_IEXEC)
```

shpc/settings.yml
```diff
@@ -50,6 +50,12 @@ docker_shell: /bin/sh
 # shell for test.sh file
 test_shell: /bin/bash
 
+# for container aliases, use wrapper shellscripts instead of shell aliases, defaults to false
+wrapper_scripts: false
+
+# shell for wrapper shellscripts
+wrapper_shell: /bin/bash
```
vsoch, line 57: Can we please put this under the other shells so they are grouped together?

shpc/main/modules/templates/singularity.lua
```diff
@@ -71,6 +76,8 @@ if (myShellName() == "bash") then
 {% endfor %}
 end{% endif %}
 
+{% endif %}
```
vsoch, line 79: Let's clean up extra white space / newlines if possible!

shpc/main/modules/__init__.py
```diff
@@ -147,7 +147,8 @@ def add(self, sif, module_name, **kwargs):
         module_name = self.add_namespace(module_name)
         template = self._load_template(self.templatefile)
         modulefile = os.path.join(self.settings.module_base, module_name.replace(":", os.sep), self.modulefile)
-        self.container.add(sif, module_name, modulefile, template, **kwargs)
+        wrapper_template = self._load_template('.'.join([self.container.templatefile,'sh']))
```
vsoch, line 150: I'm trying to think - given that we have other kinds of wrapper scripts - how would this be extended (e.g. right now self.container.templatefile is hard coded).

shpc/main/container/docker.py
```diff
@@ -214,6 +215,26 @@ def install(
         # If there's a tag in the name, don't use it
         name = name.split(":", 1)[0]
 
+        # Option to create wrapper scripts for commands
+        module_dir = os.path.dirname(module_path)
```
vsoch, line 219: The section here and below look hugely redundant! Can we not have a shared function for the container base to add wrapper scripts?

vsoch (commented):
Okay some early comments are below! Sorry for the delay, busy week and had some other shpc things to work on. Given that I tested outside of a container I think this would have worked, but it's led me to realize two things:

 - we need to figure out how to not hard code paths into the modules (and make them un-usable given a bind or change of location).
 - we should likely load the default config first and then update with the user to be able to better handle issues like forgetting to set a value.

And of course when this is more ready we need some good docs to describe wrapper scripts and show how they work and what happens when you set to true!

vsoch:
okay and here is a quick proposal to handle the bug I ran into here (not adding the wrapper_shell) https://github.com/singularityhub/singularity-hpc/pull/475

marcodelapierre:
thanks @vsoch ! should be able to work on SHPC later today

marcodelapierre:
OK, I will deep dive on this one tomorrow, for now just replying to this very first comment of yours:

> what other functionality could be provided? E.g., what would be a different kind of wrapper script, or even one that is custom provided by a user?

Great point!
Indeed, the script opens up to various possibilities, the first few that come to my mind being:
1. additional features provided directly by SHPC might include adding pre- (and eventually post- ..?) execution lines in the wrapper script. One use I can think of is for defining environment variables; this is done already in other ways in SHPC, and know I can't think of why one should use this method instead, but I am just thinking based on in-house scripts I did use
2. interface for users to edit the template (powerful and risky at same time of course)..and eventually even the possibility of having a user define template in the dot directory in the home, same as the settings. This would actually render 1. not needed...and I also now realise it's exactly what you suggested...! my brain is done for the day I guess eheh :) 

For point 2. (ie your point) then, it would be about providing similar features to `shpc config edit`/`shpc config inituser`/`shpc config edit -c`.

More coming tomorrow..

shpc/main/modules/templates/singularity.tcl
marcodelapierre: addressed in https://github.com/marcodelapierre/singularity-hpc/commit/6466461228b9c2d1d6afd62ace444c863ee3e98f

shpc/settings.yml
marcodelapierre: addressed in commit https://github.com/marcodelapierre/singularity-hpc/commit/028a1d6c1e382e36f5b9180f56d4ef6e5715eb10

shpc/main/modules/templates/singularity.lua
marcodelapierre: addressed in commit https://github.com/marcodelapierre/singularity-hpc/commit/6466461228b9c2d1d6afd62ace444c863ee3e98f

shpc/main/schemas.py
```diff
@@ -124,11 +124,13 @@
     "environment_file": {"type": "string"},
     "default_version": {"type": "boolean"},
     "enable_tty": {"type": "boolean"},
+    "wrapper_scripts": {"type": "boolean"},
```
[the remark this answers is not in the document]
marcodelapierre: addressed by @vsoch  in #475

shpc/utils/fileio.py
marcodelapierre: added in commit https://github.com/marcodelapierre/singularity-hpc/commit/0d433265cba7d6700559105391ec15a64737e44d

shpc/main/modules/templates/singularity.tcl
marcodelapierre: This is a really interesting insight, I had never encountered so far this type of scenario, in which software paths in a deployment may change. This is because most often in HPC systems, the software deployment is done and set in stone. 
I think it's something worth tackling!

How about the following ? (just using a meta language to convey the idea):
```
if SHPC_MODULE_ROOT is defined  
    set PATH to "$SHPC_MODULE_ROOT/bla-bla-bla/bin"
otherwise 
    set PATH to "hardcoded path"
```
The 1st case is for when you need the flexibility, which you can achieve by means of an env. variable.
The 2nd case is for the scenario where the deployment is set in stone.

Thoughts?

shpc/main/modules/templates/singularity.tcl
marcodelapierre: also tested

shpc/main/modules/templates/singularity.lua
marcodelapierre: also tested

shpc/utils/fileio.py
marcodelapierre: also tested

marcodelapierre:
OK, I think I am done for today.

I have provided and tested fixes for some of your feedback.

Just summarising from the above what seem to be the current open points, which I can go back to next week:
1. dynamic path in the modulefiles for wrapper scripts: feel free to give feedback to my idea, and I will then proceed with an implementation
2. flexible templatefile in the codebase
3. function to add wrapper scripts
4. further extension of the feature (for this PR, or a later one): we were talking about allowing users to provide a custom templatefile. Do you think that providing functionalities similar to the settings YAML is enough? (namely CLIs to edit the template, create a user-specific one, and edit the latter)
5. documentation for the wrapper script functionality

shpc/main/modules/templates/singularity.tcl
vsoch: I love the idea, but how do we derive the root? I was poking around LMOD and environment modules docs and couldn't figure it out. Should we ping someone with expertise?

vsoch:
> 1. dynamic path in the modulefiles for wrapper scripts: feel free to give feedback to my idea, and I will then proceed with an implementation

I love this idea and think it's so important - in practice I'm not sure how to implement it, because the module files are sort of "written once and then done." I was trying to look for a way to get this relative path via the module software but maybe there is a better way?

> 2. flexible templatefile in the codebase

This is an interesting idea that I'd like to think through more.

> 3. function to add wrapper scripts

Can you expand on this?

> further extension of the feature (for this PR, or a later one): we were talking about allowing users to provide a custom templatefile. Do you think that providing functionalities similar to the settings YAML is enough? (namely CLIs to edit the template, create a user-specific one, and edit the latter)

I think for now having a variable for a custom file would be perfect. E.g., use the default (the container ones) or write a custom one and use it. I think if we have more than one what we ultimately will want is an organized folder with named files, and then perhaps a default, but a way for an install to request using one or more (probably via the container.yaml) and then it could be provided with the container for a one off script relevant to it, or somewhere else in shpc for more generic templates. Given install of a module and request for a name, we'd first search the container.yaml directory, and then go to the shpc install.

> documentation for the wrapper script functionality

100% !

marcodelapierre:
Sorry @vsoch , been busy, will be back on it soon :)

vsoch:
okay we should be able to rebase with main and then pick up discussion here!

marcodelapierre:
Yep, in my to do list for today! ;)

shpc/main/modules/templates/singularity.tcl
marcodelapierre: In my proposition above, I was thinking SHPC_MODULE_ROOT would be defined manually by the sys admin, for those specific cases that are needed.

I gave some extra thought about this point, and in the end these wrapper scripts are similar in essence to the container SIFs in SHPC (and actually I now realise they should sit in a sub-dir of the container dir, more on this later): you hard code containerPath in the module, and can't change that, and similar story goes for the wrapper scripts.

If it's fine to have the container path hardcoded, I think it should be similarly fine to have the wrapper path hardcoded.

In addition, though, as I said, I realise the wrappers should sit in a `bin/` under the container path; in fact, under the module path they risk slowing module loading (a bit like the containers were doing in that other ticket by some other user of SHPC).
But then, if container and wrapper are in the same root directory... we could really leverage the syntax I was proposing above, in case you need to make things flexible: having a SHPC_CONTAINER_ROOT that is defined when needed, to make finding container/wrappers dynamic at loading time!

thoughts?

vsoch: That’s a good idea! I think what we could do is take any setting and allow it to be set in the environment with SHPC_<NAME> and then have an envar that falls back to the hard coded path.

I do wonder however if there is still some way for the module software to derive it? I haven’t found anything so far but maybe we can ask @xdelaruelle !

marcodelapierre: So, this does the relative path trick for Lmod!
```
prepend_path("PATH", pathJoin(string.gsub(myFileName(), "/module.lua", ""), "bin" ) )
```

marcodelapierre: and for TCL the right module variable should be `ModulesCurrentModulefile`, now just have to write the substitution function in TCL 
(be back later)

marcodelapierre: without hard-coding the module filename (Lmod):
```
prepend_path("PATH", pathJoin(myFileName():match("(.*[/])") or "./", "bin") )
```

marcodelapierre: and here is the Tcl version!

```
prepend-path PATH "[file dirname ${ModulesCurrentModulefile}]/bin"
```

marcodelapierre: these will do, provided the aliases sit in a subdir of where the modulefile is (or in the same dir, for what matters).
will work more on the code tomorrow hopefully

vsoch: You're amazing! I tried looking for both of those and came up short for both. This is super great! :partying_face:

marcodelapierre: ...and now it's in the PR, just committed! 🎉

shpc/main/modules/__init__.py
marcodelapierre: ok, first very simple pass of abstraction here,
I have defined an attribute to the `ModuleBase` class, called `wrappertemplatefile`, so that now the assignment above reads:
```
wrapper_template = self._load_template(self.wrappertemplatefile)
```
Still specific, but a bit cleaner.
In subsequent passes, selection of the template file may become more articulate

marcodelapierre:
Ok, so the big win of the day is the dynamic path in the modulefiles for wrapper scripts!
Hopefully tomorrow I will give a pass to refactoring some repeating code into a function (as per a comment of yours above).

shpc/main/container/docker.py
marcodelapierre: Ok, with the latest commit ( cf5b0c37783203463985a371a5b06c1d29befb16 ) I have created a new function , `_generate_wrapper_scripts`, in `shpc/main/container/base.py`, that gets called by the singularity and docker specific methods.

I am going to resolve this conversation and comment on the new bit of code

shpc/main/container/base.py
```diff
@@ -175,5 +178,28 @@ def get_features(self, config_features, settings_features, extra=None):
 
         return features
 
+    def _generate_wrapper_scripts(self, wrapper_template, aliases, module_dir, features, command=None, container_sif=None, image=None, tty=None):
```
marcodelapierre, line 181: @vsoch , so here it is, the new function `_generate_wrapper_scripts` to modularise the code that generates wrapper scripts. It gets called by the singularity- and docker- specific `install` methods.

Please have a look at this bit of code (especially the interface) and the corresponding method calls, to be honest I don't write OO code in Python often, so I don't have a good eye for best practices, nor conventions specific to this project :)

marcodelapierre:
Ok, so, after the modularisation of the code that generates the wrapper scripts, I can still see two outstanding points from our thread above.

The 1st one is about allowing for custom template files for the wrapper scripts. 
I am proposing to leave this feature out of this PR and work on this later (time permitting). To be entirely honest, it's really because my time is crunching right now, and I would love to see the basic wrapper scripts go public :).
I would also add that this is inline with the fact that at present there is no customisation allowed for the Lua and Tcl templates neither.

In terms of what we could do in the future for these custom templates, as we said previously, there are few possibilities:

1. one custom template (could sit in the dotdir, similar to the user settings.yml?): if the sys admins need to customise the template, they can just do it in the one that ships with SHPC, so it might make sense to offer the custom one for users (although I can't think of an immediate use for this right now)

2. custom templates that are container specific: these would go in the registry along side the container.yaml. No immediate use right now neither, but maybe there could be specific use cases.

If you're happy with keeping this in the backlog, I can turn this comment in a separated github issue.

marcodelapierre:
The 2nd outstanding point is documenting the new wrapper scripts feature, which I will try to do asap! :-)

marcodelapierre:
On a separate note, a couple of days ago I also did a commit to this PR, to ensure wildcard arguments are "quoted" appropriately in the wrapper templates, a bit like a did in the modulefile templates the other day.

Look at the weird syntax that csh needs inside a script:
```
{% if '/sh' in wrapper_shell or '/bash' in wrapper_shell %}"$@"{% elif '/csh' in wrapper_shell %}$argv:q{% endif %}
```

The wildcard is `$argv:q`, where one must use `$argv` instead of `$*`, and where `:q` takes care of possible quotes inside the argument .... this was wild! It took time me quite some time to dig this information in the internet, mostly because it was not clear to me which were the keywords to search. A couple of references, so that this doesn't get lost again:

https://stackoverflow.com/questions/42263202/bash-csh-echo-the-strings-argv-and-bin-csh
https://www.staff.tugraz.at/reinfried.o.peter/unix/cshell.html

shpc/main/container/base.py
vsoch: Let me think about this - we are hitting a pattern of "here is a template and I want to render all the things into it" and my worry is that if different wrapper scripts are added, we aren't making it easy to do that, optionally supplying custom variables to one but not the other. I do think we want a different class to describe a wrapper script, and then the main install functions will just do:

```python
from shpc.main.wrappers import generate_wrappers

if self.settings.wrapper_scripts:
    generate_wrappers(...)
``` 
And we'd provide all the general settings / config / container / paths to this function, and each wrapper template could deal with them as follows (e.g., for this first example you are requiring aliases but I don't see why all wrappers need that).

The design also needs to take into account that an install could have more than one wrapper or we might want to add a wrapper to an existing install:

```bash
$ shpc wrapper add <name> <module>
```
And then there would be a wrapper base class with a standard set of functions (that can be subclassed if needed) and we would have a master lookup.

So - this is a lot of work I suspect for a PR to ask, and I'm not even sure it will work, so if it's okay with you I can pull your branch and take a first shot.

vsoch:
> The 1st one is about allowing for custom template files for the wrapper scripts. I am proposing to leave this feature out of this PR and work on this later (time permitting). To be entirely honest, it's really because my time is crunching right now, and I would love to see the basic wrapper scripts go public :).

Let me take a shot at this, and I'll work on your branch here (but won't push in case you don't like the changes of the work!) I think it's important to have this in, for the first shot.

> I would also add that this is inline with the fact that at present there is no customisation allowed for the Lua and Tcl templates neither.

This is actually an interesting idea I was pondering - if the lua / tcl templates could be considered under the same namespace as wrapper, and then just with a different output directory. 

And I like all your ideas for custom templates! I'm hoping I can figure out something that the user can provide one, one off, and the container-specific ones make sense (if there is a use case).

marcodelapierre:
I agree with what you say, especially around the template/wrapper customisation benefitting Lua/Tcl templates, and hence the need for a wrapper class...or should it be called a template class, to effectively include lua/tcl modulefile templates? 

Please go ahead and feel free to work on this branch/PR!

On my side, I expect to shoot one more commit today with some basic documentation on the feature, and then be quite for a bit ;)

shpc/main/container/base.py
vsoch: What I'm trying to wrap my head around now is how to generalize, given that singularity.sh and docker.sh seem to be tied up tightly with the lua/tcl templates.

vsoch:
Hold off on the docs - I can do that fairly quickly, and with the refactor I'm thinking it would be not the best use of your time. I can't promise to get the PR in tonight because it's late, but I have a good start so it will be sometime this week!

marcodelapierre:
Ok sure, will wait on the docs!
And no rush on the refactoring :-)

shpc/main/container/base.py
marcodelapierre: Adding some of my thoughts, too, maybe they can be good input in the design of the new class.

The new wrapper scripts are an alternative way to the shell aliases/functions, for exposing container functionalities to users.

Some of the key differences I am aware of:
* wrapper scripts: can be called by other processes, in particular `mpirun` (mandatory for MPI applications), and scheduler managers such as Slurm `srun` (again mandatory for MPI, but also recommended for any application in a Slurm cluster); shell aliases/functions will throw an error if used this way
* shell aliases/functions: self-contained in the module, therefore allowing for a neater, more concise, more manageable deployment of the user interface to the containers; wrapper scripts on the other hand are files sitting on the filesystem (not a huge  deal, but still something to be aware of)

On the other hand, in terms of customisability, I think wrappers and shell aliases are quite similar; we realised the opportunity of customisability when working on the former, but I think that almost any edit that is allowed by the wrappers can also be implemented in the form of aliases.

In this regard, having more flexibility with both wrapper templates and tcl/lua module templates matches the customisability that is inherent in both.
Of course, module templates have even more room for customisation, as they contain more than just the command aliases.

So, that's probably why in my mind the two are quite tied up. That's also why I see the benefit of having a dedicated class, as there's quite some generality that can be expressed here.

However, I am also wondering whether there can be other types of templates or wrappers, that may expand this picture even further...? 
One thing that came to my mind several times are container-specific additional wrapper templates, that apply to specific commands. For instance, in a Rstudio container, a specific wrapper for rstudio-server (or whatever it's called), to fully customise the spawning of an Rstudio server instance (or same for Jupyter hub). 
In other words, a registry directory for a container might end up having one or all of the following files: container.yaml (mandatory), docker.sh (custom docker template for any command), singularity.sh (custom singularity template for any command), <command>-docker.sh (template for a specific <command>), <command>-singularity.sh.
Note I have been thinking out loud in the last paragraph, not sure yet whether the corner case is worth any extra work in the code.

That's all for now!

vsoch: Yes, I'm still awake! Being very bad staying up :) I have most of the work done and will PR shortly.

vsoch:
Added with #496 !

marcodelapierre:
And....I can't wait to see all of this deployed system-wide on Setonix, our upcoming Tier-1 Supercomputer here at Pawsey Centre!

PR #474, opened 2022-01-21
Create container.yaml

KoHaku55:
I create container.yaml for the specific version of star

### DIFF
--- a/registry/vclab/star/container.yaml
+++ b/registry/vclab/star/container.yaml
@@ -4,6 +4,6 @@ maintainer: 'KoHaku55'
 description: Spliced Transcripts Alignment to a Reference.
 latest:
   star2.7.2a: sha256:1f333e245e359eb7e9d73e798423a346be50cb620c1a8e4784366fae2e021a67
-tag:
+tags:
 aliases:
   STAR: /usr/local/bin/STAR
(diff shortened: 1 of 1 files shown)


registry/vclab/star/container.yaml
```diff
@@ -0,0 +1,8 @@
+docker: tipsarin/star2.7.2a
+url: https://github.com/vclabsysbio/AI-MD_RNASeq_ChIPseq/blob/0f2ddcddc38f5d0eaa5b1712d467482b2d0a1ba5/STAR/
+maintainer: 'KoHaku55'
+description: Spliced Transcripts Alignment to a Reference.
+latest:
```
vsoch, line 5: This might be redundant but you’ll also need a tags section here (with latest) to update if there are ever other versions.

vsoch: Confirmed in CI because recipe did not validate. 

03ADC780-3CEE-41BE-BB03-94E267251EC2

vsoch:
Tiny detail - tags should have the same latest and sha.

I'm going back to bed but will finish review tomorrow!

registry/vclab/star/container.yaml
```diff
@@ -0,0 +1,10 @@
+docker: tipsarin/star2.7.2a
```
vsoch, line 1: So if we want to be consistent with namespace, then the module folder for this container should be:

```bash
registry/tipsarin/star2.7.2a/container.yaml
```
E.g., if someone had the namespace of "vclab" on docker hub you shouldn't be able to claim it here. I also think that is why the test is failing - it's looking for vclab/star/ as a module (but per the container name it should match that).

muffato:
I would vote to close this pull-request as there is already a container.yaml for this software, _star_, added in #409, and automatically updated to the latest version by the cron job, e.g. #478

vsoch:
Agree @muffato - and it's been 5 months without response. @KoHaku55 if you have any questions or want to add a variant (as the current star doesn't support your needs) feel free to re-open / open an updated PR.

PR #475, opened 2022-01-22
Ensuring that we always load defaults and then update with user settings

vsoch:
Right now, when shpc is updated with new config values and the user has a custom setting, you can trigger an error if a needed value is missing. to get around this, we should always load the default settings file first, and then override with custom settings given that the preferred settings are not == the default


vsoch:
@marcodelapierre do you have some time to review? This is related to #473

marcodelapierre:
thanks for pinging @vsoch ! should be able to work on SHPC later today

marcodelapierre:
did a quick test, it works, and it makes complete sense to me!

vsoch:
Thank you!

PR #476, opened 2022-01-26
testing parsing port

vsoch:
This is a WIP - it likely will break some interactions down the line so we need someone that has a container registry with port to test!


vsoch:
Closing for now as not wanted- can re-address in the future if needed.

PR #481, opened 2022-02-03
Update singularity.lua

surak:
No need to set SINGULARITY*_OPTS them here - they come from the shell script or not.


vsoch:
This looks reasonable! And I think you want to do the same for singularity.tcl, docker.tcl, and docker.lua that also set them. And please bump the version in shpc/version.py and add the corresponding change to CHANGELOG.md in the root.

Pinging @marcodelapierre and @xdelaruelle because if I remember correctly we set them for a reason and I want to make sure we aren’t messing anything up! 😆

vsoch:
Why did you close?

surak:
> Why did you close?

See #483

PR #482, opened 2022-02-03
Update use-cases.rst

surak:
The text assumes that the container clearly has the `--nv` set, which was not the case on `nvcr.io/nvidia/pytorch/21.11-py3`, and it doesn't tell which variable should be set for adding it.

### DIFF
--- a/docs/getting_started/use-cases.rst
+++ b/docs/getting_started/use-cases.rst
@@ -196,7 +196,7 @@ Adding Options
 --------------
 
 By default, some of the commands will come with singularity options. For example,
-a container intended for gpu should have the ``--nv`` flag. However,
+a container intended for gpu will have a feature: gpu set to true, and this will add the ``--nv`` flag given that the user or cluster settings file has that feature enabled. However,
 it could be the case that you want to define custom options at the time of use.
 In this case, you can export the following custom environment variables to add them:
 
--- a/docs/getting_started/use-cases.rst
+++ b/docs/getting_started/use-cases.rst
@@ -196,12 +196,12 @@ Adding Options
 --------------
 
 By default, some of the commands will come with singularity options. For example,
-a container intended for gpu is always going to give you the ``--nv`` flag. However,
+a container intended for gpu should have the ``--nv`` flag. However,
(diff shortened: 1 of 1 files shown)


docs/getting_started/use-cases.rst
```diff
@@ -196,12 +196,12 @@ Adding Options
 --------------
 
 By default, some of the commands will come with singularity options. For example,
-a container intended for gpu is always going to give you the ``--nv`` flag. However,
+a container intended for gpu should have the ``--nv`` flag. However,
```
vsoch, line 199: @surak I don’t think tensorflow is one of the containers flagged for gpu?

vsoch: I think probably in your settings.yaml you didn’t turn on gpu so it wasn’t installed with it - can we check that?

surak: No, but PyTorch is -it's an nvidia one and it was compiled with gpu support and yet, it's not flagged as such. In any case, if one brings a container from elsewhere, one won't know that the module should be flagged, right?

vsoch: If PyTorch needs the features gpu flag we can add it. But the center or user still needs to set gpu to true in the settings.yaml.

vsoch: See here: https://github.com/singularityhub/singularity-hpc/blob/bfb2c33796d60e291639e0725f00b3bcff084aa7/shpc/settings.yml#L63

vsoch, line 199: How about this?
```suggestion
a container intended for gpu will have a feature: gpu set to true, and this will add an the ``--nv`` flag given that the user or cluster settings file has that feature enabled. However,
```

surak: Fine with me!

vsoch: Hmm interesting - so the hope was that there would be two controllers - one for the center / user to be able to say "always add these flags given that the container says they are supported (the settings here) and then the "on the fly" add this flag given that the center hasn't done that. I do think the current change of the PR is much better and I'd like to merge!

PR #485, opened 2022-02-04
bugfix for conflict() in modulefile templates 4x

marcodelapierre:
Hi @vsoch ,

have a look at these edits, I think they improve the way conflicts for modules are declared in the modulefile templates.

Bottom line: therefore, the main conflict was with either myModuleName() or {{ name }}; I think a more appropriate one is a having both {{ name }} and {{ tool }}. Plus, the conditional for alias names (checked against {{ tool }} ).

Example for quay.io/biocontainers/fastqc:0.11.9--0
- BEFORE: either "quay.io/biocontainers/fastqc:0.11.9--0" or "quay.io/biocontainers/fastqc"
- NOW: both "quay.io/biocontainers/fastqc" and "fastqc"
    The latter is useful for conflicts with other, non SHPC, modules.

### DIFF
--- a/shpc/main/modules/templates/docker.lua
+++ b/shpc/main/modules/templates/docker.lua
@@ -54,7 +54,7 @@ local inspectCmd = "{{ command }} ${PODMAN_OPTS} inspect ${PODMAN_COMMAND_OPTS} 
 set_shell_function("{|module_name|}-shell", shellCmd,  shellCmd)
 
 -- conflict with modules with the same name
-conflict("{{ tool }}"{% if aliases %}{% for alias in aliases %}{% if alias.name != tool %},"{{ alias.name }}"{% endif %}{% endfor %}{% endif %})
+conflict({% if name != tool %}"{{ name }}",{% endif %}"{{ tool }}"{% if aliases %}{% for alias in aliases %}{% if alias.name != tool %},"{{ alias.name }}"{% endif %}{% endfor %}{% endif %})
 
 -- exec functions to provide "alias" to module commands
 {% if aliases %}{% for alias in aliases %}
--- a/shpc/main/modules/templates/docker.lua
+++ b/shpc/main/modules/templates/docker.lua
@@ -38,7 +38,7 @@ For each of the above, you can export:
 setenv ("PODMAN_OPTS", "")
 setenv ("PODMAN_COMMAND_OPTS", "")
 
--- we probably don't need this
+-- we probably do not need this
 local MODULEPATH="{{ module_dir }}"
 
 -- interactive shell to any container, plus exec for aliases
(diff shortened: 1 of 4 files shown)


shpc/main/modules/templates/docker.lua
```diff
@@ -54,7 +54,7 @@ local inspectCmd = "{{ command }} ${PODMAN_OPTS} inspect ${PODMAN_COMMAND_OPTS}
 set_shell_function("{|module_name|}-shell", shellCmd,  shellCmd)
 
 -- conflict with modules with the same name
-conflict(myModuleName(){% if aliases %}{% for alias in aliases %}{% if alias.name != name %},"{{ alias.name }}"{% endif %}{% endfor %}{% endif %})
+conflict({% if name != tool %}"{{ name }}",{% endif %}"{{ tool }}"{% if aliases %}{% for alias in aliases %}{% if alias.name != tool %},"{{ alias.name }}"{% endif %}{% endfor %}{% endif %})
```
vsoch, line 57: I don't see how this is a better fix? Where does tool come from (note it could be undefined if it's set here https://github.com/singularityhub/singularity-hpc/blob/efb3ff13c653455525a54e2cc0e90111e567c34b/shpc/main/modules/__init__.py#L44) and wouldn't it be an issue if the module name is mypython and that's provided by another module (so we would just want to conflict the module name regardless as we do now). And also the if logic is changed so there will be a hanging comma if no aliases.

shpc/main/modules/templates/docker.tcl
```diff
@@ -51,8 +51,9 @@ set helpcommand "This module is a {{ docker }} container wrapper for {{ name }}
 {% endfor %}{% endif %}
 
 # conflict with modules with the same alias name
-conflict {{ name }}
-{% if aliases %}{% for alias in aliases %}{% if alias.name != name %}conflict {{ alias.name }}{% endif %}
+{% if name != tool %}conflict {{ name }}{% endif %}
```
vsoch, line 54: I see the same issue here - I don't think we need to check if name !- tool.

shpc/main/modules/templates/docker.lua
marcodelapierre: I think that here tool is unrelated to the naming convention that can be set in the settings file.

It seems to me that `tool` comes from the template rendering call:   https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/container/singularity.py#L185

and hence from `parsed_name.tool` . Here, `parsed_name` seems to originally come out of the attributes of the ContainerName class:   https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/container/base.py#L15-L38

Putting more comments below..

marcodelapierre:
Let me recap @vsoch , here are what seem to me possible token that could be used for module conflicts, with example values from 3 containers:

|  | myModuleName() | name | tool |

vsoch (approved):

PR #486, opened 2022-02-04
bugfix for module templates: consistency of aliases and passed arguments

marcodelapierre:
Hi @vsoch ,

so I incurred into this one while testing a production Nextflow workflow (https://github.com/marcodelapierre/illumina-nf) with SHPC.

The bad guy was this command:
```
blast_formatter \
    -archive blast_contigs_sub.asn \
    -outfmt "6 qaccver saccver pident length evalue bitscore stitle" \
    -out blast_unsort_contigs_sub.tsv
```

and the bad thing about this bad guy are those double quotes.

So I had a look at the modulefile templates, and protected the argument placeholders `$@` with escaped double quotes, `\"$@\"`, for the command aliases.

That made it!

Tested the fixed command aliases it with both LMOD and TCL. Note the issue only actually arises with bash, because csh does not require any placeholder.

Now, just an extra note, I was also able to fix this issue for `|module|-exec` and `|module|-run`, but ... only for LUA, not for TCL. 
Why? The problem is that LUA allows to provide distinct syntaxes for bash and csh, whereas TCL does not seem to.
So, I left the TCL templates as before, without placeholders for those `-exec` and `-run` commands.


vsoch (commented):
I feel like we've gone back and forth on this, and we cannot just remove the extra argument to indicate parsing command line args from the user. I don't see how this matches to quotes escaping?

marcodelapierre:
Ok, I have worked a bit more on this one, too, with (I hope) some important improvements stemming from your first feedback.

The PR now covers 4 issues, all related to alias generation and their arguments.

1. Quotes around the argument placeholder for bash (all 4x templates). Here is the thing, thanks to the quoted arguments, `"$@"`, the following alias will always work:
    ```
    $ type python
    python is a function
    python () 
    { 
        singularity ${SINGULARITY_OPTS} exec ${SINGULARITY_COMMAND_OPTS} -B /group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/singularity-hpc/../modules/python/3.9.2-slim/99-shpc.sh:/.singularity.d/env/99-shpc.sh /group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/singularity-hpc/../containers/python/3.9.2-slim/python-3.9.2-slim-sha256:ce367bb30b8928efb632c369e3bd4a8dbd7905417bd0a245087a82c250e54a24.sif /usr/local/bin/python "$@"
    }
    ```
    On the other hand, without the double quotes the alias will fail, if the passed arguments happens to contain double quotes.
    My first change in the templates tackles this issue.

2. Handling bash vs csh in definition of `-exec` and `-run` aliases (2x TCL templates only): similar to the command aliases, I have added a conditional depending on the shell, so that the argument placeholder is added for bash but not for csh, as it should be, and as already implemented for the LUA case.

3. Fixing command variable in 2x Docker templates: the alias command was referred as `alias.args`, which does not exist; I have fixed it to be `alias.command`

4. Fixing non-existing (and probably not needed) ELSE conditional in help output (2x Singularity templates only): in the output message, there was an ELSE conditional for when no aliases are defined in the recipe, which would have created a standard `|module|` alias, identical to `|module|-run`; this was in the help, but then not implemented in the modulefile. Because it's the same as `|module|-run`, I have gotten rid of the help message, rather than implementing it. However, I could do the latter if you prefer.

Hope this clarifies a bit the sense of this PR :)

shpc/main/modules/templates/docker.lua
```diff
@@ -25,7 +25,7 @@ Commands include:
        {{ command }} inspect <container>
 
 {% if aliases %}{% for alias in aliases %} - {{ alias.name }}:
-       {{ command }} run -i{% if tty %}t{% endif %} -u `id -u`:`id -g` --rm --entrypoint {{ alias.entrypoint }} {% if envfile %}--env-file {{ module_dir }}/{{ envfile }}{% endif %} {% if bindpaths %}-v {{ bindpaths }} {% endif %}{% if features.home %}-v {{ features.home }} {% endif %}{% if alias.docker_options %}{{ alias.docker_options }} {% endif %} -v ${PWD} -w ${PWD} <container> "{{ alias.args }}"
+       {{ command }} run -i{% if tty %}t{% endif %} -u `id -u`:`id -g` --rm --entrypoint {{ alias.entrypoint }} {% if envfile %}--env-file {{ module_dir }}/{{ envfile }}{% endif %} {% if bindpaths %}-v {{ bindpaths }} {% endif %}{% if features.home %}-v {{ features.home }} {% endif %}{% if alias.docker_options %}{{ alias.docker_options }} {% endif %} -v ${PWD} -w ${PWD} <container> "{{ alias.command }}" "$@"
```
vsoch, line 28: I'm wondering why this works - e.g.,:

- the full entrypoint + args starts as a value, this is alias.command
- we split this value into args.entrypoint and args.args (see here https://github.com/singularityhub/singularity-hpc/blob/7b844f15b89d09431e3eb5ed54abc0ee171aa92d/shpc/main/container/__init__.py#L194-L198)
- so what we are doing now is having the entrypoint PLUS repeating it for the exec command at the end. Does this really work? And to step back - why wasn't the entrypoint being honored before? I am thinking maybe we need to use alias.command for singularity, but maintain alias.entrypoint -> entrypoint and alias.args for the doker/podman templates?

shpc/main/modules/templates/docker.lua
```diff
@@ -60,7 +60,7 @@ conflict(myModuleName(){% if aliases %}{% for alias in aliases %}{% if alias.nam
 
 -- exec functions to provide "alias" to module commands
 {% if aliases %}{% for alias in aliases %}
-set_shell_function("{{ alias.name }}", execCmd .. {% if alias.docker_options %} "{{ alias.docker_options }} " .. {% endif %} " --entrypoint {{ alias.entrypoint }} " .. containerPath .. " {{ alias.args }} $@", execCmd .. {% if alias.docker_options %} "{{ alias.docker_options }} " .. {% endif %} " --entrypoint {{ alias.entrypoint }} " .. containerPath .. " {{ alias.args }}")
+set_shell_function("{{ alias.name }}", execCmd .. {% if alias.docker_options %} "{{ alias.docker_options }} " .. {% endif %} " --entrypoint {{ alias.entrypoint }} " .. containerPath .. " {{ alias.command }} \"$@\"", execCmd .. {% if alias.docker_options %} "{{ alias.docker_options }} " .. {% endif %} " --entrypoint {{ alias.entrypoint }} " .. containerPath .. " {{ alias.command }}")
```
vsoch, line 63: This also sets `--entrypoint` so it would be like doing:


```bash
$ docker exec --entrypoint /bin/bash <container> /bin/bash "@"
```
and I think we still want

```bash
$ docker exec --entrypoint /bin/bash <container> "@"
```

shpc/main/modules/templates/docker.tcl
```diff
@@ -74,24 +76,32 @@ set-alias {|module_name|}-shell "${shellCmd}"
...
 } else {
-{% for alias in aliases %}  set-alias {{ alias.name }} "${execCmd} {% if alias.docker_options %} {{ alias.docker_options | replace("$", "\$") }} {% endif %} --entrypoint {{ alias.entrypoint | replace("$", "\$") }} ${containerPath} {{ alias.args | replace("$", "\$") }}"
+{% for alias in aliases %}  set-alias {{ alias.name }} "${execCmd} {% if alias.docker_options %} {{ alias.docker_options | replace("$", "\$") }} {% endif %} --entrypoint {{ alias.entrypoint | replace("$", "\$") }} ${containerPath} {{ alias.command | replace("$", "\$") }}"
 {% endfor %}
 }
 {% endif %}
 
 # A customizable exec function
-set-alias {|module_name|}-exec "${execCmd} --entrypoint \"\" ${containerPath}"
+if { [ module-info shell bash ] } {
```
vsoch, line 93: This makes sense! And I remember the empty entrypoint was needed for it to work.

shpc/main/modules/templates/singularity.lua
```diff
@@ -16,7 +16,7 @@ Container:
 Commands include:
 
  - {|module_name|}-run:
-       singularity run {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home }} --home {{ features.home }} {% endif %}{% if features.x11 %}-B {{ features.x11 }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container>
+       singularity run {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home }} --home {{ features.home }} {% endif %}{% if features.x11 %}-B {{ features.x11 }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container> "$@"
```
vsoch, line 19: It looks like here we didn't remove bindpaths in the command / or it wasn't rebased with current main / or I forgot to remove here :P

shpc/main/modules/templates/singularity.lua
```diff
@@ -27,8 +27,8 @@ Commands include:
...
-{% endfor %}{% else %} - {|module_name|}: singularity run {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home }} --home {{ features.home }} {% endif %}{% if features.x11 %}-B {{ features.x11 }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }}{% endif %}<container>{% endif %}
+       singularity exec {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home }} --home {{ features.home }} {% endif %}{% if features.x11 %}-B {{ features.x11 }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}{% if alias.singularity_options %}{{ alias.singularity_options }} {% endif %}<container> {{ alias.command }} "$@"
```
vsoch, line 30: okay - alias.command makes sense here because we want the full entrypoint + args (there is no `--entrypoint`)

vsoch (commented):
This is looking really good! I'm about to go to bed but wanted to comment so you wouldn't wait for me to sleep :) 

Adding the quotes to escape the args seems spot on - but I'm wondering if replacing args with command is really the right solution, as the command include an entrypoint plus args, and for docker/podman we are already defining the entrypoint. So I think that needs to be changed back, although I haven't tested it. With bash it would essentially be running bash in the container as entrypoint to start another bash shell.

> Handling bash vs csh in definition of -exec and -run aliases (2x TCL templates only): similar to the command aliases, I have added a conditional depending on the shell, so that the argument placeholder is added for bash but not for csh, as it should be, and as already implemented for the LUA case.

Perfect.

> Fixing command variable in 2x Docker templates: the alias command was referred as alias.args, which does not exist; I have fixed it to be alias.command

I think this is incorrect, per my reason above. If an alias doesn't have args it will be empty, and the actual alias entrypoint is passed via `--entrypoint`.

> Fixing non-existing (and probably not needed) ELSE conditional in help output (2x Singularity templates only): in the output message, there was an ELSE conditional for when no aliases are defined in the recipe, which would have created a standard |module| alias, identical to |module|-run; this was in the help, but then not implemented in the modulefile. Because it's the same as |module|-run, I have gotten rid of the help message, rather than implementing it. 

That makes sense!

> Hope this clarifies a bit the sense of this PR :)

Yes! It's coming along really nicely, thank you for fixing these things up.

vsoch:
Just doing some testing - I think if we want to be consistent and use command, we can set the entrypoint to empty as we do later, e.g., this seems to work!

```bash
$ docker run -it --entrypoint "" ubuntu /usr/bin/echo "HELLO"
```

shpc/main/modules/templates/singularity.lua
marcodelapierre: confirm, the -B was not removed yet in the help outputs :)

marcodelapierre:
on args vs command, let me do a quick check, maybe I need indeed to be reverse this

shpc/main/modules/templates/singularity.lua
vsoch: Oh you know what, I did that on purpose so the user would see they are still getting the bind.

marcodelapierre:
On command vs args:

... indeed,  I had forgotten about the entrypoint, that's why I thought there was something to be fixed

So, I reverted that specific commit, and update the description of the PR

shpc/main/modules/templates/docker.lua
marcodelapierre: you're right, `args` is the right one!! :)

shpc/main/modules/templates/docker.lua
marcodelapierre: you're right, `args` is the right one!! :)

shpc/main/modules/templates/singularity.tcl
```diff
@@ -18,19 +18,19 @@ proc ModulesHelp { } {
     puts stderr "Commands include:"
     puts stderr ""
     puts stderr " - {|module_name|}-run:"
-    puts stderr "       singularity run {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home | replace("$", "\$") }} --home {{ features.home | replace("$", "\$") }} {% endif %}{% if features.x11 %}-B {{ features.x11 | replace("$", "\$") }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container>"
+    puts stderr "       singularity run {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home | replace("$", "\$") }} --home {{ features.home | replace("$", "\$") }} {% endif %}{% if features.x11 %}-B {{ features.x11 | replace("$", "\$") }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container> \"$@\""
```
vsoch, line 21: Do we not need to escape the `$` for these two changed lines here? (e.g., the command at the end).

vsoch: Note there are two - one here and below!

vsoch, line 21:
```suggestion
    puts stderr "       singularity run {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home | replace("$", "\$") }} --home {{ features.home | replace("$", "\$") }} {% endif %}{% if features.x11 %}-B {{ features.x11 | replace("$", "\$") }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container> \"\$@\""
```

shpc/main/modules/templates/singularity.tcl
```diff
@@ -18,19 +18,19 @@ proc ModulesHelp { } {
...
     puts stderr " - {|module_name|}-exec:"
-    puts stderr "       singularity exec {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home | replace("$", "\$") }} --home {{ features.home | replace("$", "\$") }} {% endif %}{% if features.x11 %}-B {{ features.x11 | replace("$", "\$") }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container> $*"
+    puts stderr "       singularity exec {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home | replace("$", "\$") }} --home {{ features.home | replace("$", "\$") }} {% endif %}{% if features.x11 %}-B {{ features.x11 | replace("$", "\$") }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container> \"$@\""
```
vsoch, line 25:
```suggestion
    puts stderr "       singularity exec {% if features.gpu %}{{ features.gpu }} {% endif %}{% if features.home %}-B {{ features.home | replace("$", "\$") }} --home {{ features.home | replace("$", "\$") }} {% endif %}{% if features.x11 %}-B {{ features.x11 | replace("$", "\$") }} {% endif %}{% if envfile %}-B {{ module_dir }}/{{ envfile }}:/.singularity.d/env/{{ envfile }}{% endif %} {% if bindpaths %}-B {{ bindpaths }} {% endif %}<container> \"\$@\""
```

vsoch: And just curious - why did this work before without the quote and escape?

marcodelapierre: edited - 3x in the help output of both docker and singularity tcl templates.

good question - with modules 3.2.10.6 it seems to work either ways, with no change in the output rendering...
I added the escaping as per your suggestion, seems the safe option

vsoch (approved):
okay, this looks consistent! Let's merge and I'll do a release.

PR #491, opened 2022-02-05
adding release workflow

vsoch:
This will allow me to make a release on GitHub, and then trigger a release to pypi.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,29 @@
+name: Release Python Package
+
+on:
+  release:
+    types: [created]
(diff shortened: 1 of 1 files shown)


vsoch:
Let's test this!

marcodelapierre:
github and pypi release versions in sync: I think this can be quite important for reproducible deployments. 

Even if you pip install shpc, you will still need to clone the registry from github, and so it's good to always have the option to keep versions in sync. This is only one of the possible ways people may choose to enforce reproducibility control for their software deployments, and one which I think is good to provide.

PR #494, opened 2022-02-14
avoid double bindpath: use -B, ditch SINGULARITY_BINDPATH

marcodelapierre:
hi @vsoch , so I liked your point on the readability of using the variable (see #489), so I attempted to keep it, and use prepend_path/prepend-path , by also specifying the comma delimiter.

Unfortunately, this syntax didn't work with neither Lmod or Modules, considering I am using slightly older versions: Lmod 7.6.1 and Modules 3.2.10.6.
I guess this may be the case also in other computing centres, and it's not something that gets updated often if this is the case.

So I ended up re-adding `-B` and dropping the variable.

### DIFF
--- a/shpc/main/modules/templates/singularity.lua
+++ b/shpc/main/modules/templates/singularity.lua
@@ -41,8 +41,7 @@ For each of the above, you can export:
 -- we probably don't need this
 local MODULEPATH="{{ module_dir }}"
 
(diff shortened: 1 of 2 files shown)


vsoch (approved):
Perfecto! Let’s merge into main (no release) and do a release when we get the other changes in.

PR #496, opened 2022-02-16
Add/wrapper scripts

vsoch:
this will support two kinds of wrapper scripts - global that live in the settings.yaml, and container.yaml specific that go alongside that file. We basically will then generate aliases for any executable names that were not created as wrappers. we will want to follow this up with a command to list wrappers, and also include descriptions for the interested user, and a flag to install to select on the fly
   
cc @marcodelapierre ! I hope this is getting closer to what we want - being able to support custom wrappers for aliases (the global ones) and alongside containers. It is HUGELY more complicated than your original implementation (and this might be bad) but on the other hand, adding a new wrapper is fairly trivial.Take a look and let me know what you think - it's dangerous for me to stay up so late, and who knows what monster I coded! On the other hand, it was totally worth it because I had a ton of fun!

I also tried to clean up some of the initial module parsing, e.g., instead of passing self.settings variables a million times I'm just using self.settings, and i tried to do that for wrappers too.


vsoch:
oy vey! So I added a bug it seems. I'm going to need your help to debug the module templates again @marcodelapierre, sorry!

vsoch:
Looks like when I first pushed with wrappers enabled there was a different error too:

```
Testing load of python:3.9.5-alpine
/usr/local/Modules/init/bash: line 88: alias: -s: invalid option
alias: usage: alias [-p] [name[=value] ... ]
Test of python was not successful.
['python']
```

vsoch:
okay I think the issue was the tool name came from settings, and I updated this to use parsed_name.<tool> so it was rendering an alias name starting with `-s` or `-e` which is invalid.

marcodelapierre:
Hi @vsoch , sorry have been unwell for the past couple of days, so still mostly off the screen to rest (excuse me if anything below is confusing to read).

I feel a bid guilty in reading that you spent a night coding this one up....but glad you had fun in doing it! :-)

So just an overall comment right now, your feature overview reads amazing!
> this will support two kinds of wrapper scripts - global that live in the settings.yaml, and container.yaml specific that go alongside that file. We basically will then generate aliases for any executable names that were not created as wrappers.

I really think it's a winning idea to provide custom templates in the form of container specific templates that target specific commands. I think it also hits the target "audience" most likely to use it, i.e. developers of yaml recipes.

Just a couple of comments as I had a look at your new developer documentation, but not haven't looked at the implementation.

So, I understand there's an amazing fine grained of using general/specific wrapper scripts (Singularity as an example here):
* wrapper_scripts: false - all disabled
* wrapper_scripts: true AND singularity: null  - only container specific scripts enabled (if found)
* wrapper_scripts: true AND singularity: singularity.sh - generic scripts with standard template, plus container specific scripts enabled (if found) 
* wrapper_scripts: true AND singularity: <custom path> - generic scripts with custom template, plus container specific scripts enabled (if found) 

This is brilliant, I love it!!


And another one: from the developer guide, it seems like the `bin/` directory for wrapper scripts is under the SIF directory, whereas the comment in settings.yml suggests it's under the module directory.

Which one is true? From a quick look, it seems like the implementation puts it under the containers dir. 
Just beware that this would give troubles with our relative path trick in the modulefile templates, which I notice is still there. You should then change the modulefile templates so that the prefix is the absolute container path where the SIF is found.

As much as I like my original relative path discoveries in lua/tcl languages, I think the container subdir may be better from a functionality point o view: in the implementation where the scripts are under the module dir/, the `module` commands still have to scan them to understand whether they're legal modulefiles or not, which can be time consuming. (although I had run a test with the quay.io bbmap container, with 200 wrapper scripts, and saw no measurable difference)

Last question: are there any specific points you want me to check/debug?
I will run some stock tests any way...when I am back in full capacity.

vsoch:
Oh no worries! I hope you feel better soon :) I had a stressful week too, kind of glad it's closing up!

```
    wrapper_scripts: false - all disabled
    wrapper_scripts: true AND singularity: null - only container specific scripts enabled (if found)
    wrapper_scripts: true AND singularity: singularity.sh - generic scripts with standard template, plus container specific scripts enabled (if found)
    wrapper_scripts: true AND singularity: - generic scripts with custom template, plus container specific scripts enabled (if found)
```
Yep! Just one adjustment - if someone writes 

```
    wrapper_scripts: true AND singularity: - generic scripts with custom template, plus container specific scripts enabled (if found)
```
It's going to look for a file named `singularity` in the present working directory and shpc/wrappers, and if it doesn't find it, will error. People can more reasonably put a fullpath to a file.

> And another one: from the developer guide, it seems like the bin/ directory for wrapper scripts is under the SIF directory, whereas the comment in settings.yml suggests it's under the module directory.

Oops that's just a typo - I'll take a look after this. We do want it under containers to not slow down module load.

> Just beware that this would give troubles with our relative path trick in the modulefile templates, which I notice is still there. You should then change the modulefile templates so that the prefix is the absolute container path where the SIF is found.

Oh oops, I didn't think about this. Hmm. Should we change it to the module directory and take the hit for a little bit of slowness (assuming files sparse and small)  or change to use fullpath to containers?

> As much as I like my original relative path discoveries in lua/tcl languages, I think the container subdir may be better from a functionality point o view: in the implementation where the scripts are under the module dir/, the module commands still have to scan them to understand whether they're legal modulefiles or not, which can be time consuming. (although I had run a test with the quay.io bbmap container, with 200 wrapper scripts, and saw no measurable difference)

Hmm with 200 wrappers?! So maybe putting alongside modules isn't such a terrible idea?

> Last question: are there any specific points you want me to check/debug? I will run some stock tests any way...when I am back in full capacity.

I think just some discussion about where to put the scripts, and then you can test!

marcodelapierre:
So, on the location of the scripts, `<containers-path>/bin`, vs `<modules-path>/bin`, as both implementations seem technically doable, I have the question, where do you think these scripts conceptually belong to?

I personally tend to see them as part of the actual package (although created separately by SHPC), as they are file objects that provide an interface to access the raw executables. 
The main reason why I had still put them in the module dir in my original implementation, was the case of docker/podman containers: the containers dir is not needed for their container images, and I was unsure whether it was worth to keep that dir only for the wrapper scripts. I hadn't mentioned this point to you. 
Oh, this also implies that we need to define and pass the information to template.render on the theoretical (i.e. defined but not used for the image) `container_path` also for docker/podman containers.

So, if these don't sound too ugly to you, I would go for this one. But this is just one view. :)

vsoch:
One thing I just thought of - so for either container or module - we would be okay with the user changing container mid install and over-writing the old scripts? 

I do think if you don't have container binaries, the container directory is excessive, and we could just do the module directory. That might be reason enough for me!

So "go for this one" you mean stick with module directory? I am definitely okay with that!

vsoch:
And if we go with module dir, no more changes are needed (e.g., this is correct?)
```
{% if wrapper_scripts %}prepend-path PATH "[file dirname ${ModulesCurrentModulefile}]/{{ settings.wrapper_subdir }}"{% endif %}
```
lordy I'm looking at my code from that "stay up all night session" - I don't know this person who wrote all that :laughing:

vsoch:
Oh I see and I need to change:
```
        # Write scripts into container directory
        wrapper_dir = os.path.join(
            self.container_dest_dir, self.settings.wrapper_subdir
        )
```
to use the module_dir

vsoch:
okay I think I did it!
```
modules/
├── python
│   └── 3.9.10
│       ├── 99-shpc.sh
│       ├── bin
│       │   └── python
│       └── module.lua
```
will push asap.

marcodelapierre:
good! 

Yes, I remember the feeling of being surprised at the contents of code written in late night sessions eheh.

doing some testing now, then

marcodelapierre:
ok, little thing to fix here, I am having a problem in changing the `enabled` setting:
```
md@zeus-1:wrap-vsoch$ shpc config set wrapper_scripts:enabled:True
wrapper_scripts:True cannot be added to config: 'True' is not of type 'boolean'
md@zeus-1:wrap-vsoch$ shpc config set wrapper_scripts:enabled:true
wrapper_scripts:true cannot be added to config: 'true' is not of type 'boolean'
md@zeus-1:wrap-vsoch$ shpc config set wrapper_scripts:enabled:false
wrapper_scripts:false cannot be added to config: 'false' is not of type 'boolean'
md@zeus-1:wrap-vsoch$ shpc config set wrapper_scripts:enabled:False
wrapper_scripts:False cannot be added to config: 'False' is not of type 'boolean'
```

other settings work fine.

now ahead with (un)installation tests

vsoch:
okay should be fixed: https://github.com/singularityhub/singularity-hpc/pull/496/commits/e73e16e7d2eb4e5addcbd78f49d37954b8e0ffd5

marcodelapierre:
ok thanks.

another few issues:

1. in the singularity module templates, there are some leftover occurrences of `{{ singularity_shell }}`, that have to become `{{ settings.singularity_shell }}` ; note 
2. the x11 feature does not produce the extra strings in the modules/wrappers, for a silly reason: when setting x11 to true using `shpc config set ..` , the new value is in single quotes, and hence interpreted as of type string instead of boolean

on point 2., a possibly related behaviour: I have noticed that the single quotes are added by `shpc config set` every time the starting value is null. For instance:
```
shpc config set wrapper_scripts:singularity:null
shpc config set wrapper_scripts:singularity:singularity.sh
```
will result in a final settings.yml which has a single quoted string value.

marcodelapierre:
several singularity tests, OK

now docker

marcodelapierre:
docker OK


found one little issue (singularity, but I guess generic):

this causes an error (wrappers enabled, but generic singularity set to false - to only use custom wrappers in the recipe dir):
```
wrapper_scripts:
  enabled: true

  # use for docker aliases (set to null to disable)
  docker: docker.sh

  # use for podman aliases (set to null to disable)
  podman: docker.sh

  # use for singularity aliases (set to null to disable)
  singularity: null
```

```
$ shpc install quay.io/biocontainers/blast:2.12.0--pl5262h3289130_0
  File "/group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/wrap-vsoch/singularity-hpc/bin/shpc", line 11, in <module>
    load_entry_point('singularity-hpc', 'console_scripts', 'shpc')()
  File "/group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/wrap-vsoch/singularity-hpc/shpc/client/__init__.py", line 372, in run_shpc
  File "/group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/wrap-vsoch/singularity-hpc/shpc/client/install.py", line 17, in main
    container_tech=args.container_tech,
  File "/group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/wrap-vsoch/singularity-hpc/shpc/main/__init__.py", line 28, in get_client
    settings = Settings(kwargs.get("settings_file"), validate)
  File "/group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/wrap-vsoch/singularity-hpc/shpc/main/settings.py", line 280, in __init__
    self.validate()
  File "/group/pawsey0001/mdelapierre/PSS/shpc-tests/wrap-18jan22/wrap-vsoch/singularity-hpc/shpc/main/settings.py", line 50, in validate
    jsonschema.validate(instance=self._settings, schema=shpc.main.schemas.settings)
  File "/group/pawsey0001/mdelapierre/.local/lib/python3.6/site-packages/jsonschema/validators.py", line 934, in validate
    raise error
jsonschema.exceptions.ValidationError: None is not of type 'string'

Failed validating 'type' in schema['properties']['wrapper_scripts']['properties']['singularity']:
    {'type': 'string'}

On instance['wrapper_scripts']['singularity']:
    None

```

marcodelapierre:
And a very last comment, this is a proposed change for discussion.

Currently, in modulefile templates we're using the dynamic module_dir inspection only to define the PATH for wrapper scripts. 
However, we're still using a `{{ module_dir }}` in other instances:
* modulefile templates: to provide the location of the environment file
* modulefile templates: to define a service variable (MODULEPATH for Lmod, progdir for Tcl)
* wrapper templates: to provide the location of the environment file

I think we should be able resort to a dynamic definition in all these cases, which would make the module tree completely portable, the only (unavoidable probably) hard-coded path remaining the SIF path for Singularity.
The only non trivial one could be in the wrapper templates, if people use a strange subdir instead of `bin/` (as in, different directory tree level), although there's probably a solution for this one (?).

Thoughts?

vsoch:
@marcodelapierre I think I've fixed the null parsing (and related validation) issues! For the module_dir I agree, and if you want to make suggestions on the PR here or just write in comments how to change, I can work on tomorrow! You can also PR to the branch here and then I'll just merge it in. I'm off to bed now but thank you for the review - back tomorrow!

marcodelapierre:
Ok, can confirm all of the reported issues above are resolved when using the latest commit.

Two minor polishing edits:
* `settings.yml`, `wrapper_subdir` value, change `"bin"` into `bin` (no quotes) for consistency with other string values
* `modules/templates/docs.md` : typos regarding `shell` vs `settings.singularity_shell`

I have actioned these two edits in a branch of mine, doing the PR onto your branch now ..

marcodelapierre:
For `module_dir`, I am unable to edit the code myself now (possibly in the next few days..?), but I was thinking:

- module templates: use the dynamic module_dir syntax to define a service variable, which could be called `moduleDir` for instance, and then use this variable to build the following:
  - PATH for wrapper scripts
  - path of environment file to bind mount
- by the way, we could unify `MODULEPATH` (Lua) and `progdir` (Tcl) to something like `moduleDir` as said above, or if there is any reason for these names, we will then keep using them

- wrapper templates: 
  - first of all, it would be good to add a couple of checks when parsing the value of `settings.wrapper_subdir` in SHPC, including removing leading/trailing slashes, and possibly only allowing one level of subdirectory (i.e `bin` is fine, but `sub/bin` to be made illegal). We should also disallow climbing back in the tree with `..`, which would mess up things across container versions and even repositories, and we might want to also disallow the current directory `.` , just to keep things tidy and easy. If we do all of this, then it's super simple to have the dynamic path. For instance in bash: 
  ```
  wrapperDir="$(dirname $(readlink -f $0) )"
  moduleDir="${wrapperDir}/../"
  ```

Thoughts?

vsoch:
> first of all, it would be good to add a couple of checks when parsing the value of settings.wrapper_subdir in SHPC, including removing leading/trailing slashes, and possibly only allowing one level of subdirectory (i.e bin is fine, but sub/bin to be made illegal). We should also disallow climbing back in the tree with .., which would mess up things across container versions and even repositories, and we might want to also disallow the current directory . , just to keep things tidy and easy. 

Totally agree - and another basic question - is there any need to really allow them to customize bin? Like what else would they change it to? Then we don't have any of those problems about parsing paths / checking for going up in the tree!

marcodelapierre:
Good point! I think there's no need to expose the choice of the `bin` string to admins/users, it could just be defined as an attribute of a WrapperScript object.

I can see only two occurrences of `wrapper_subdir` outside the templates. 
They are inside `wrappers/base.py`, and with an object attribute they could just become: `self.settings.wrapper_subdir` --> `self.wrapper_subdir`

vsoch:
okay done! I didn't even keep the variable I just put bin, we can go back to a variable when we know we need it!

marcodelapierre:
ok, so two outstanding bits:
1. merge #497 in here, with some minor polishings
2. module_dir changes, see above

on point 2., for the wrapper scripts better to keep it as simple as:
* bash
    ```
    moduleDir="$(dirname $0)/.."
    ```
* csh
    ```
    set moduleDir="`dirname $0`/.."
    ```
* template for both shells (inverted quotes work in bash, too):
    ```
    {% if '/csh' in settings.wrapper_shell %}set {% endif %}moduleDir="`dirname $0`/.."
    ```

vsoch:
@marcodelapierre before I push any changes - what about:

```
script=`realpath $0`
wrapper_bin=`dirname $script`
{% if '/csh' in settings.wrapper_shell %}set moduleDir=`dirname $wrapper_bin`{% else %}export moduleDir=$(dirname $wrapper_bin){% endif %}
```
And then we can have that as a snippet added to the wrappers at the top?

marcodelapierre:
Hi @vsoch , sure , reads good to me.
feel free to go ahead with all these module_dir changes if you have the time, including in the lua/tcl templates.
I was hoping to go ahead myself soon, but am unable to.

vsoch:
@marcodelapierre okay I think I had a cool idea! So I was thinking we will likely want to add snippets/bases for different wrapper scripts, so I refactored the directory to organize as such (see docs in https://github.com/singularityhub/singularity-hpc/pull/496/commits/4c7ac07a627af3a7a4772f95f2b8101baa452b8a) and now the scripts can do things like extend or include other templates!

This doesn't apply for modules, but I'm thinking if we like the approach and there is huge redundancy in the tcl/lua templates we could do something similar (but for another PR).

Also I wasn't sure what you wanted me to do for tcl/lua templates with the bash/shell so I will need your guidance there.

marcodelapierre:
No, it's only a problem if the quotes are there (in my case it was from an old settings.yml).

`shpc config set` now correctly writes true/false without quotes.
So, maybe mine is a non-issue..? :)

vsoch:
I would say it’s a non issue if it’s impossible to do again, but if we can be more strict in the schema to not allow it in the first we could try that. I think this PR is getting huge so my thinking is this issue is slightly out of scope especially if we fixed the setting to prevent it! But if someone hits it again we can come back to it.

marcodelapierre:
OK done:

https://github.com/singularityhub/singularity-hpc/pull/499

marcodelapierre:
If I am not missing something...I think this PR might be ready, wow! :-) 
What do you think @vsoch  ?
(no rush, I am just excited!)

vsoch:
I agree! :partying_face: 

Will merge and release after dinner.

PR #499, opened 2022-03-02
Add/wrapper scripts

marcodelapierre:
Edits to use a dynamically defined `moduleDir` variable in all templates, rather than fixed `{{ module_dir }}`:
* wrapper templates: location of env-file
* modulefile templates
  * defining a local moduleDir variable
  * used for wrappers PATH, location of env-file, help output
  * both LUA and TCL
* no more need to pass `module_dir` when rendering modulefile templates

Fix in wrapper base script:
* for CSH shell, variables always need to be defined with `set `
* note that `export` is not needed for bash shell, as the variable is only used directly in the wrapper script, not in any sub process
  * if willing to revert, then for consistency the CSH case needs `setenv` (can't recall the exact syntax)


vsoch:
okay looking good! I think there are two more places that we missed changing module_dir:

- https://github.com/singularityhub/singularity-hpc/blob/add/wrapper-scripts/registry/vanessa/salad/docker_fork.sh
- https://github.com/singularityhub/singularity-hpc/blob/add/wrapper-scripts/registry/vanessa/salad/singularity_fork.sh

marcodelapierre:
edits now done for `module_dir` in `vanessa/salad/`!

PR #500, opened 2022-03-02
fixing version updates for some quay.io/biocontainers

marcodelapierre:
Some automatic updates to tag versions for quay.io/biocontainers did not catch the real latest version.
Here I am manually fixing these for some containers.
I haven't investigated the reason in the code.

Note that these biocontainers seem to have the following tagging convention, that follows that of bioconda:
<package-version>-<some-alphanum-string>_<rebuild-index>


vsoch:
@marcodelapierre the filters look ok - we only run the updates once a month could that be it? If they push new containers nightly or weekly that would explain the difference.

marcodelapierre:
it is really strange, because in most cases, the newly picked tag is indeed older in time than the previously existing ones..

vsoch:
okay issue opened - sorry for the bug @marcodelapierre ! If we merge here the correct updated versions should last at least a few weeks https://github.com/autamus/cuppa/issues/3

vsoch (approved):

Issue #505, opened 2022-03-08
How to enable wrapper scripts ?


muffato:
Hello,

It seems that 
```diff
diff --git a/shpc/settings.yml b/shpc/settings.yml
index b02344346..9816ecad3 100644
--- a/shpc/settings.yml
+++ b/shpc/settings.yml
@@ -60,7 +60,7 @@ wrapper_shell: /bin/bash
 # If you want a custom wrapper script for a one-off container, define it there with the script
 # stored alongside the container.yaml.
 wrapper_scripts:
-  enabled: false
+  enabled: true
 
   # use for docker aliases (set to null to disable)
   docker: docker.sh
```
is not sufficient to get the wrapper scripts, as I get this error message:
```
$ shpc install ruby:3.0.2
singularity pull --name /lustre/scratch123/tol/teams/tolit/users/mm49/shpc/singularity-hpc/containers/ruby/3.0.2/ruby-3.0.2-sha256:15dd21ae353c5f4faebed038d9d131c47b9fd84c14be8c3cfbc750204b63f009.sif docker://ruby@sha256:15dd21ae353c5f4faebed038d9d131c47b9fd84c14be8c3cfbc750204b63f009
INFO:    Using cached SIF image
INFO:    Using cached SIF image
/lustre/scratch123/tol/teams/tolit/users/mm49/shpc/singularity-hpc/containers/ruby/3.0.2/ruby-3.0.2-sha256:15dd21ae353c5f4faebed038d9d131c47b9fd84c14be8c3cfbc750204b63f009.sif
Traceback (most recent call last):
  File "/nfs/users/nfs_m/mm49/.conda/envs/shpc/bin/shpc", line 33, in <module>
    sys.exit(load_entry_point('singularity-hpc', 'console_scripts', 'shpc')())
  File "/lustre/scratch123/tol/teams/tolit/users/mm49/shpc/singularity-hpc/shpc/client/__init__.py", line 372, in run_shpc
    main(args=args, parser=parser, extra=extra, subparser=helper)
  File "/lustre/scratch123/tol/teams/tolit/users/mm49/shpc/singularity-hpc/shpc/client/install.py", line 19, in main
    cli.install(args.install_recipe)
  File "/lustre/scratch123/tol/teams/tolit/users/mm49/shpc/singularity-hpc/shpc/main/modules/__init__.py", line 332, in install
    self.container.install(
[...]

Issue #504, opened 2022-03-08
shpc pulls images twice


muffato:
Hello !

I'm testing the `add/symlink-install` branch and noticed something again that's bugging me. Whenever I install a container, shpc always pulls it twice. Here is a "screenshot" from using fc4f684be68c88b47f716352bb27fba1b425d456. I don't have any local modifications.

```
$ shpc install ruby:3.0.2
singularity pull --name /lustre/scratch123/tol/teams/tolit/users/mm49/shpc/singularity-hpc/containers/ruby/3.0.2/ruby-3.0.2-sha256:15dd21ae353c5f4faebed038d9d131
c47b9fd84c14be8c3cfbc750204b63f009.sif docker://ruby@sha256:15dd21ae353c5f4faebed038d9d131c47b9fd84c14be8c3cfbc750204b63f009
INFO:    Converting OCI blobs to SIF format
INFO:    Starting build...
Getting image source signatures
Copying blob sha256:647acf3d48c2780e00cd27bb0984367415f270d78477ef9d5b238e6ebd5290da
Copying blob sha256:b02967ef003473d9adc6e20868d9d66af85b0871919bcec92419f65c974aa8ce
Copying blob sha256:e1ad2231829e42e6f095971b5d2dc143d97db2d0870571ba4d29ecd599db62cb
Copying blob sha256:5576ce26bf1df68da60eeb5162dccde1b69f865d2815aba8b2d29e7181aeb62b
Copying blob sha256:a66b7f31b095b7fa01d8ba10e600a192bab43a1311f50216cf6fa9a45d0f435e
Copying blob sha256:11c270d8f828fe2d06ecdcab761a226d5de10060af6a34af7e3bed50159e3b7e
Copying blob sha256:5df92bd6f19a6f98e0a07e6f56f1ca22e5404393ddb728b0a7d6d130160a62ea
Copying blob sha256:834a3e3a7652bcc6309bace7a87e43a4b746f9b50eaeea799fa8a77830ba290d
Copying config sha256:2aae56d7a3a57dd63f6296371c1228a51fb00cfac10b516c85ee60841d2bb139
Writing manifest to image destination
Storing signatures
2022/03/08 17:48:57  info unpack layer: sha256:647acf3d48c2780e00cd27bb0984367415f270d78477ef9d5b238e6ebd5290da
2022/03/08 17:48:57  warn xattr{etc/gshadow} ignoring ENOTSUP on setxattr "user.rootlesscontainers"
[...]

PR #506, opened 2022-03-08
bugfix for singularity pull (repeated output) and wrapper scripts template

vsoch:
This PR will fix:

- #505 : the "templates" variable under wrapper scripts needs to be referenced differently because it will trigger an error if null (possibly there are differences based on how the yaml is parsed?)
- #504 : this was just a change to an underlying dependency, spython, and without updating here it prints twice. Updating here to correctly use stream we not only print once, but we get the lines in real time! So overall a big improvement on both fronts!


vsoch:
@muffato this is all set if you'd care to review!

muffato:
The first two modules were sufficient, but the rest is OK too 👍🏼

PR #508, opened 2022-03-09
Updated the paths for the Perl image

muffato:
Hello !
Very minor thing. When doing my tests, I found that the Perl executables in `/opt/view/bin/` are now suffixed with the version itself. This PR is simply to update `container.yml`

Cheers,
Matthieu

### DIFF
--- a/registry/ghcr.io/autamus/perl/container.yaml
+++ b/registry/ghcr.io/autamus/perl/container.yaml
@@ -5,12 +5,11 @@ description: The Perl programming language.
 latest:
   5.35.0: sha256:66602a525597110eb12d5b3a944180d9a8e7af1f02d60dba3237503f6b532959
 tags:
-  5.33.3: sha256:ddbe04d704c8883aed0959dbd88bd8410cf5e1a6775a2f46264af306723b6459
   5.35.0: sha256:66602a525597110eb12d5b3a944180d9a8e7af1f02d60dba3237503f6b532959
   latest: sha256:66602a525597110eb12d5b3a944180d9a8e7af1f02d60dba3237503f6b532959
 aliases:
-  perl: /opt/view/bin/perl
-  perlbug: /opt/view/bin/perlbug
-  perldoc: /opt/view/bin/perldoc
-  perlivp: /opt/view/bin/perlivp
-  perlthanks: /opt/view/bin/perlthanks
+  perl: /opt/view/bin/perl5.35.0
+  perlbug: /opt/view/bin/perlbug5.35.0
+  perldoc: /opt/view/bin/perldoc5.35.0
+  perlivp: /opt/view/bin/perlivp5.35.0
+  perlthanks: /opt/view/bin/perlthanks5.35.0


vsoch:
Hmm... how will we handle when there are version changes?

vsoch:
Okay looking at spack (from which the view is derived) I don't see any issues with the paths:

```
$ ls opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/perl-5.34.0-bvgnm2ejnajpvaruta22d5c24g6qi4zu/bin/
corelist  h2ph       perl        perlthanks  pod2text    ptardiff   xsubpp
cpan      h2xs       perl5.34.0  piconv      pod2usage   ptargrep   zipdetails
cpanm     instmodsh  perlbug     pl2pm       podchecker  shasum
enc2xs    json_pp    perldoc     pod2html    prove       splain
encguess  libnetcfg  perlivp     pod2man     ptar        streamzip
```
E.g., perl5.34.0 is there but perl is too! Of course that's not the same version (going to check the one we have for the container next). Indeed you are correct the autamus container has the version:

```
# ls /opt/view/bin/perl*
/opt/view/bin/perl5.35.0     /opt/view/bin/perlivp5.35.0
/opt/view/bin/perlbug5.35.0  /opt/view/bin/perlthanks5.35.0
/opt/view/bin/perldoc5.35.0
```
Ah but I just installed 5.35.0 and yep, no more non-versioed names!

```
 ls opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/perl-5.35.0-smm2auqgf4jwkvi7uzvrvjmqdxcwqtkc/bin/
corelist5.35.0   json_pp5.35.0     pl2pm5.35.0       ptardiff5.35.0
cpan5.35.0       libnetcfg5.35.0   pod2html5.35.0    ptargrep5.35.0
cpanm            perl5.35.0        pod2man5.35.0     shasum5.35.0
enc2xs5.35.0     perlbug5.35.0     pod2text5.35.0    splain5.35.0
encguess5.35.0   perldoc5.35.0     pod2usage5.35.0   streamzip5.35.0
h2ph5.35.0       perlivp5.35.0     podchecker5.35.0  xsubpp5.35.0
h2xs5.35.0       perlthanks5.35.0  prove5.35.0       zipdetails5.35.0
instmodsh5.35.0  piconv5.35.0      ptar5.35.0
```

I'm going to ask spack folks about this - at face value I'd want to be able to generate a view that honors the name of the original executable. I suspect they took this approach to allow more than one perl in the view, but if there isn't support for it, I think we should be able to turn this off.

vsoch:
https://github.com/spack/spack/issues/29409

vsoch:
okay I'm going to put in a PR to hopefully fix this, let's see if they will like my solution.

vsoch:
https://github.com/spack/spack/pull/29413

vsoch:
@muffato so the PR is under review - not sure if it will be reviewed soon or even ultimately approved! But one thing I think I learned is that we only have this "versioned binary" thing for odd numbers (which are considered development versions I think?) so an easy fix to try here would be to:

1. remove development versions from current tags (odd numbers)
2. add a filter for even numbers only.

What do you think?

muffato:
Hi @vsoch. Ah yes I should have realised that ! Yes, I remember from my Perl days that the public, stable, versions have got an even number, and the dev one an odd number.
It makes sense indeed to discard the odd numbers from the registry / autamus

vsoch:
Sounds good! here are some ideas to play with:

```
In [6]: re.search("[0-9]+[.][0-9]{1}[02468][.]*", "5.35.0")

In [7]: re.search("[0-9]+[.][0-9]{1}[02468][.]*", "5.34.0")
Out[7]: <re.Match object; span=(0, 5), match='5.34.'>

In [8]: re.search("[0-9]+[.][0-9]{1}[02468][.]*", "5.33.0")

In [9]: re.search("[0-9]+[.][0-9]{1}[02468][.]*", "5.32.0")
Out[9]: <re.Match object; span=(0, 5), match='5.32.'>
```
We'd want to add as a filter:

```
filter:
- "[0-9]+[.][0-9]{1}[02468][.]*"
```
ping @alecbcs to ask if a regular expression pattern like this will also work for binoc updates?

vsoch:
@alecbcs that would be fantastic! Do you have bandwidth to test it out or do you need me to? If you point me to the right spots and what you have in mind I can take a shot in some evening this week, but if it's pretty easy / simple and you want to jump in and do it that works too! This definitely I think is considered a bug at this point - not for this particular container (which cycles between develop and non-develop) but for others that have their hashes / latest updated to actually older versions.

PR #510, opened 2022-03-09
Minor refactoring of `check_symlink`

muffato:
Hiya,

There are two changes in this pull-request.

The first two commits are a follow up of 10cf953d9c29b282781b0e50c10bd4b3ab7295ef. No need to call `mkdirp` for `symlink_base` since `mkdirp` is anyway called in `create_symlink` to create the parent directory. Then I slightly refactor `check_symlink` to give it a similar interface to `create_symlink`, i.e.: both are only called if the user wants symlinks, and neither need to check the symlink parameter.

The last commit is about stripping `module.tcl` from being shown by `module avail` in the symlinks tree. This is done by making the symlinks point at the `module.tcl` files directly, rather than their enclosing directories. It makes the symlink tree much shorter and nicer to use.
To achieve that, I had to change the module templates to hardcode the path to the modules, because otherwise it couldn't find the `bin/` of the wrapper scripts.

Example trees:
```
containers/
├── golang
│   └── 1.18-rc
│       └── golang-1.18-rc-sha256:c9be2380bfdc0f4da95e1635e914dbd596bc73b618cdf16894a631eb58945156.sif
└── quay.io
    └── biocontainers
        ├── bwa
        │   └── 0.7.17--h7132678_9
        │       └── quay.io-biocontainers-bwa-0.7.17--h7132678_9-sha256:07822e4293a8c59755b295c448b9541db6c9bdbfdedb010bdbdcc1e1e935370f.sif
        ├── bwa-mem2
        │   └── 2.2.1--hd03093a_2
        │       └── quay.io-biocontainers-bwa-mem2-2.2.1--hd03093a_2-sha256:df046077d0771d622d2889039050a995a71430ed155f0f94f52e944bf2785b2a.sif
        ├── cooler
        │   ├── 0.8.11--pyh3252c3a_0
        │   │   └── quay.io-biocontainers-cooler-0.8.11--pyh3252c3a_0-sha256:c954780cf75fd8fe026de342253982a354b6211258e45f66e18d08af68271138.sif
        │   └── 0.8.6--py_0
        │       └── quay.io-biocontainers-cooler-0.8.6--py_0-sha256:1b0e2e9625d9d0beee5befaf866df016a80f364339fbb396ab8d84878a3a0584.sif
        └── samtools
            └── 1.15--h3843a85_0
[...]

PR #511, opened 2022-03-10
Skip `module` in the symlinks

muffato:
Hiya,

Following the comments that were raised in #510 , I've found away of making the symlinks point at the `module.tcl` directly, rather than their enclosing directory, *without* having to hardcode the module directory 💪🏼 

As a reminder, the purpose is to make the output of `module avail` shorter, e.g. from
```
bwa-mem2/2.2.1--hd03093a_2/module.tcl  cooler/0.8.6--py_0/module.tcl           golang/1.18-rc/module.tcl             
bwa/0.7.17--h7132678_9/module.tcl      cooler/0.8.11--pyh3252c3a_0/module.tcl  samtools/1.15--h3843a85_0/module.tcl  
```
to
```
bwa-mem2/2.2.1--hd03093a_2  bwa/0.7.17--h7132678_9  cooler/0.8.6--py_0  cooler/0.8.11--pyh3252c3a_0  golang/1.18-rc  samtools/1.15--h3843a85_0  
```

I've never done any Lua, so only changed the Tcl templates. The change is not as trivial as I was expecting because:
- `file normalise` doesn't normalise the filename, only the directories
- `file readlink` raises an error if called on a direct file

So, I had to use `file type` to identify the symlinks, and then call `file readlink`.

If people find this change useful, and agree with the implementation method, the Lua code would still have to be written.

I'm aware that @marcodelapierre you said you were busy. no worries !

Matthieu


muffato:
In the end, I'm simply skipping this functionality for Lmod because:
1. It breaks `default_version=false`
2. We said in https://github.com/singularityhub/singularity-hpc/pull/510#issuecomment-1062520879 that modules had to be portable, and couldn't include hard-coded paths (except the container itself). The module (and its symlink) must be able to resolve `bin/` (for the wrapper scripts) under the _real_ path. In Lmod, LUA scripts are evaluated in a sandbox that has access to a limited number of functions, cf https://github.com/TACC/Lmod/blob/master/src/sandbox.lua#L66-L260. Notably, the list excludes `io.popen`, which is typically used to resolve symlinks.

In summary, this PR removes `module.tcl` from the TCL symlinks, and leaves the LUA symlinks as they were in @vsoch 's original implementation

muffato:
Not sure how you want to deal with that, @vsoch 

This is the branch I'm using locally for all my tests. I've merged the latest `main` and #517 because they both impact / conflict with the implementation of symlinks.

For review here, there are two main commits: https://github.com/singularityhub/singularity-hpc/pull/511/commits/f469cb7613694d639f41908929db43ad3e6f6ae7 and https://github.com/singularityhub/singularity-hpc/pull/511/commits/c67172367dbcdb3b7a27e5bc3c23269c26c97e39; and https://github.com/singularityhub/singularity-hpc/pull/511/commits/2ea12ac98727fb003dc6e3f2a8e7385f65fce99b

The simplified symlink tree is now available for Lmod too, if `default_version==True`.

vsoch:
Deal with what @muffato ?

muffato:
With the review. The two merge commits make the diff harder to review

vsoch:
oh I see! So I think I just need to rebase that older PR with main (as you have done) so they don't show up as new.

vsoch:
okay @muffato I think I was able to update both PRs so the number of files changed is now 9 and not 16! Can you take a quick look and make sure I resolved the merge conflicts correctly?

muffato:
All good 👍🏼

muffato:
Added two commits:
 - no need to abort the installation if the user doesn't want the symlink to be overwritten
 - added a command-line option (`--force`) to force overwriting the symlink

shpc/main/modules/__init__.py
```diff
@@ -229,7 +229,8 @@ def check_symlink(self, module_dir):
         if os.path.exists(symlink_path) and not utils.confirm_action(
             "%s already exists, are you sure you want to overwrite?" % symlink_path
         ):
-            sys.exit(0)
+            return False
```
vsoch: Hmm I don't think I agree with this change - if the user aborts we need to stop.

shpc/main/modules/__init__.py
```diff
@@ -219,17 +219,20 @@ def create_symlink(self, module_dir):
         # Create .version
         self.write_version_file(os.path.dirname(symlink_path))
 
-    def check_symlink(self, module_dir):
+    def check_symlink(self, module_dir, force):
```
vsoch, line 222: should probably have a default here! And +1 I like this change - it makes sense to have force here to mirror other commands!

```suggestion
    def check_symlink(self, module_dir, force=False):
```

shpc/main/modules/__init__.py
```diff
@@ -219,17 +219,20 @@ def create_symlink(self, module_dir):
...
         """
         # Get the symlink path - does it exist?
         symlink_path = self.get_symlink_path(module_dir)
-        if os.path.exists(symlink_path) and not utils.confirm_action(
-            "%s already exists, are you sure you want to overwrite?" % symlink_path
-        ):
-            return False
+        if os.path.exists(symlink_path):
+            if force:
+                logger.info("Overwriting %s, as requested")
```
vsoch, line 231:
```suggestion
                logger.info(f"Overwriting {symlink_path}, as requested")
```

muffato: Went for the traditional %-formatting because there are no f-strings in shpc so far. It felt odd introducing the first one here

shpc/main/modules/__init__.py
muffato: With `sys.exit(0)` in place, the module won't be installed at all. One wouldn't be able to install say both `ruby:3.1.0` and `ghcr.io/autamus/ruby:3.1.0` at the same time. I thought that `shpc` was meant to still allow the same software+version to exist in multiple namespaces, and the user would essentially choose which one to put as symlink.

vsoch: We don't currently allow that - you can only install to one namespace (and then would be able to symlink to a name at the lowest level). 

I do think we want to exit here, given the user has explicitly said "no I don't want to overwrite," and they can decide to install the module not as a symlink. I think it's easier to press up and reproduce the install command without the symlink flag as opposed to seeing it install, needing to control C or otherwise try to undo it.

shpc/main/modules/__init__.py
vsoch: Sounds good! I've only started to start using format strings myself (for other new libraries) and i suspect at some point might do a sweep through old ones to change the style. It's funny I still prefer `%s` when the string is kind of long/gnarly or otherwise weird (e.g., index of a dictionary always is weird). Anyway +1 to this choice.

shpc/main/modules/__init__.py
muffato: Assuming people used the `--symlink-tree` command-line option. If `symlink_tree` is enabled globally in `settings.yaml`, one would have to edit the latter, install the software, and edit the config file back

vsoch: That’s a good point. But also if they are requiring symlink tree I’d argue a namespace conflict in the tree would prevent them from installing, period. I would rather see the opposite flag for this case, —no-symlink-tree, than for an install to continue that a person has explicitly said no to. Or I’d rather see a confirm choose with a numbered set of options:

1. continue install without symlink

2. abort install

muffato: I added `--no-symlink-tree`

shpc/main/modules/__init__.py
```diff
@@ -343,7 +357,24 @@ def check(self, module_name):
         config = self._load_container(module_name.rsplit(":", 1)[0])
         return self.container.check(module_name, config)
 
-    def install(self, name, tag=None, symlink=False, **kwargs):
+    def write_version_file(self, version_dir):
```
vsoch, line 360: Is this section old from before? I guess we need to merge the version changes before this one!

shpc/client/__init__.py
```diff
@@ -103,7 +103,21 @@ def get_parser():
     install.add_argument(
         "--symlink-tree",
         dest="symlink",
-        help="install to symlink tree too.",
+        help="install to symlink tree too (overrides settings.yml).",
+        default=None,
```
vsoch, line 107: Oh this is neat I’ve never seen it before! So if —symlink-tree is set, the None is True and this overrides the default False of no symlink tree (should that be Nome too)? And then if no symlink tree is True the None is ignored? What happens if both flags are provided?

vsoch: Perhaps we should use store_const here to make the three cases more clear? https://stackoverflow.com/a/34750557

muffato: > this overrides the default False of no symlink tree (should that be None too)

It works as it is: None if no option given, and True / False for `--symlink-tree` / `--no-symlink-tree`

> What happens if both flags are provided?

If both are given, the last one wins

> Perhaps we should use store_const here to make the three cases more clear? https://stackoverflow.com/a/34750557

A matter of choice :) I find `store_true` and `store_false` very explicit :)

shpc/main/modules/__init__.py
muffato: Yes, because I had done a merge of the `tcl_default_version` :/
I can probably remove this merge commit to make this PR clearer ?

shpc/client/__init__.py
vsoch: Ok, I’m cool with that :)

will probably finish review tomorrow - I want to look everything over once more with a clear head, and I’m rather tired and still a bit on edge from the day. So stay tuned for tomorrow! I think the default version PR looks great (and can update the PR here) so probably I’ll merge that one first.

goodnight from the inferno! lol

F1103B3A-C907-4872-91DA-C70F9447F509

shpc/main/modules/__init__.py
vsoch: Yes - let's update this PR with the changes in current main and then I can give another review!

marcodelapierre:
Just a note to remember keeping an eye on #502, which is related to this one.
@muffato and @vsoch, you have the bigger picture I think, from what I can tell the missing bit, with proposed implementation in that other PR, is the implementation of the short naming for symlinks also for Lmod.

vsoch:
@muffato just to follow up here - I think we want to see some kind of tweak /rebase so these version changes don't show up here anymore (that are now deprecated in favor of the other PRs that we worked on!) I'm not sure the best way to go about it if it's not trivial to fix, but let me know what you'd suggest.

muffato:
Hiya. I've got on my local checkout a version of this branch, with the latest `main` merged in, the conflicts solved, and all tested on the HPC. I'd suggest pushing my branch, and changing the target branch of this PR to `main`. The diff would then essentially show the content of #502 + the present changes for removing `module`. Then we approve + merge this PR here, and can forget about #502 (actually, it depends how you would do the merge. With an actual merge commit, GitHub would automatically close #502 because it would see the commits following each-other. But I think you tend to a squash + rebase ? In that case, I'm not sure GitHub could find out)

vsoch:
oh gotcha! That does make sense. So I'm good to merge into the symlink branch (which is against main) and we can pick up from there.

PR #516, opened 2022-03-16
Miscellaneous changes

muffato:
Hi,
For your consideration: a few minor changes that I came to implement when working on symlinks and default versions, but that don't strictly relate to any of these features.
Best,
Matthieu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,6 @@ __pycache__
 *.simg
 *.sif
 *.img
-.eggs
-symlinks
-modules
+/.eggs
+/symlinks
+/modules
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ shpc.db
 singularity_hpc.egg-info/
 env
 .env
-env
 build
 docs/_build
 release
--- a/shpc/main/modules/__init__.py
+++ b/shpc/main/modules/__init__.py
@@ -104,14 +104,12 @@ class ModuleBase(BaseClient):
         """
         Sub function, so we can pass more than one folder from uninstall
         """
-        if os.path.exists(module_dir) and not force:
-            msg = "%s, and all content below it? " % name
-            if utils.confirm_uninstall(msg, force):
-                self._cleanup(module_dir)
-                logger.info("%s and all subdirectories been removed." % name)
-
-        elif os.path.exists(module_dir) and force:
-            shutil.rmtree(module_dir)
+        if os.path.exists(module_dir):
+            if not force:
+                msg = "%s, and all content below it? " % name
+                if not utils.confirm_uninstall(msg, force):
+                    return
+            self._cleanup(module_dir)
             logger.info("%s and all subdirectories have been removed." % name)
         else:
             logger.warning("%s does not exist." % name)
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -4,7 +4,6 @@ shpc.db
 singularity_hpc.egg-info/
 env
 .env
-env
```
vsoch, line 7: The instructions show creating an env like this so it doesn’t hurt to have (and that’s my preference).

shpc/main/modules/__init__.py
```diff
@@ -104,17 +104,15 @@ def _uninstall(self, module_dir, name, force=False):
...
+        if os.path.exists(module_dir):
+            if not force:
+                msg = "%s, and all content below it? " % name
+                if not utils.confirm_uninstall(msg, force):
+                    return
             shutil.rmtree(module_dir)
             logger.info("%s and all subdirectories have been removed." % name)
         else:
-            logger.info("%s does not exist." % name)
+            logger.warning("%s does not exist." % name)
```
vsoch, line 115: This works! Could you confirm you see the message in mustard yellow when it doesn’t exist? It makes me suspicious why I chose info instead of warning!

vsoch (commented):
These look great! A few minor things then if you can bump the version and CHANGELOG we can release.

.gitignore
muffato: Removing this line because `env` is already listed two lines above. Unless I'm misreading it ?

vsoch: Rofl. No I’m just blind don’t mind me 🤓

shpc/main/modules/__init__.py
muffato: Screenshot 2022-03-16 at 02 44 03

vsoch: Perfect thank you.

vsoch:
Okay so the last bit of housekeeping is to bump the version in shpc/version.py and then add a corresponding line in CHANGELOG.md for it.

shpc/main/modules/__init__.py
```diff
@@ -104,17 +104,15 @@ def _uninstall(self, module_dir, name, force=False):
...
-            if utils.confirm_uninstall(msg, force):
-                self._cleanup(module_dir)
-                logger.info("%s and all subdirectories been removed." % name)
-
-        elif os.path.exists(module_dir) and force:
+        if os.path.exists(module_dir):
+            if not force:
+                msg = "%s, and all content below it? " % name
+                if not utils.confirm_uninstall(msg, force):
+                    return
```
vsoch, line 111: Sorry just missed one thing - what happened to:

```
self._cleanup(module_dir)
```
The subtle difference is cleaning up the module base:

https://github.com/singularityhub/singularity-hpc/blob/c9eca19b80d6f59a2ab216d8f93ac9cb2ccc2b3b/shpc/main/modules/__init__.py#L59-L63

Here a module directory might be the <container>/<version> however if all the versions are deleted, we'd want to move up one level and delete the container folder too. Take that one level higher, we clean up the modules directory.

muffato: Ah, right ! Should this line https://github.com/singularityhub/singularity-hpc/blob/c9eca19b80d6f59a2ab216d8f93ac9cb2ccc2b3b/shpc/main/modules/__init__.py#L114 have been `self._cleanup(module_dir)` in the first place ?

vsoch: Yep, or at least now I can’t think of why the two cases should be different beyond my own oversight!

PR #517, opened 2022-03-16
Implemented "default_version==false" for TCL

muffato:
Following the discussion in #502, I played with `.version` and the trick I mentioned in https://github.com/singularityhub/singularity-hpc/pull/502#issuecomment-1065813278
Unfortunately it doesn't work with Lmod, but since it works in TCL, I decided to implement it, and essentially expand the `default_version` option to TCL.
It's a bit awkward, because for Lmod, `.version` has to be created when the option is turned on, and for TCL, `.version` has to be created when the option is turned off 🙃 !

Even tough the Lmod one is an empty file, I've still put both under `templates/`, and shpc copies the right one according to the options. The code is already in a function, `write_version_file`, so that it can be called in #502 from the symlink code.

Demo (TCL + `default_version=false`):
```
$ shpc install quay.io/biocontainers/samtools:1.15--h3843a85_0; shpc install quay.io/biocontainers/samtools:1.14--hb421002_0
Module quay.io/biocontainers/samtools:1.15--h3843a85_0 was created.
Module quay.io/biocontainers/samtools:1.14--hb421002_0 was created.

$ tree -a modules/
modules/
└── quay.io
    └── biocontainers
        └── samtools
            ├── 1.14--hb421002_0
            │   ├── 99-shpc.sh
            │   └── module.tcl
            ├── 1.15--h3843a85_0
            │   ├── 99-shpc.sh
            │   └── module.tcl
            └── .version

$ module load quay.io/biocontainers/samtools/
ERROR: Unable to locate a modulefile for 'quay.io/biocontainers/samtools/please_specify_a_version_number'
```

And for Lmod + `default_version=false`, no `.version` is created, and:
```
$ module load quay.io/biocontainers/samtools
Lmod has detected the following error:  The following module(s) are unknown: "quay.io/biocontainers/samtools"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "quay.io/biocontainers/samtools"

Also make sure that all modulefiles written in TCL start with the string #%Module
```

Regards,
Matthieu


vsoch (changes requested):
oops forgot to hit the submit button here! This is a great start, and I have a few suggestions / requests for changes. Let me know if you want to talk anything through further!

shpc/main/modules/templates/no_default_version.tcl
```diff
@@ -0,0 +1,2 @@
+#%Module
+set ModulesVersion "please_specify_a_version_number"
```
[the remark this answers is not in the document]
muffato: No, that was not the intent. This is just a way of preventing loading a module without specifying the version number, like what Lmod does by default.
What you're suggesting (a way of defining for each software what the default version should be, instead of letting Lmod/TCL do it based on its own rules) is a different feature, that imho should be spec'ed and addressed in a different PR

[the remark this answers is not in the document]
muffato: It works this way (TCL):
```
$ module load samtools/=
ERROR: Unable to locate a modulefile for 'samtools/please_specify_a_version_number'
```

comparing to Lmod:
```
$ module load samtools
Lmod has detected the following error:  The following module(s) are unknown: "samtools"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "samtools"

Also make sure that all modulefiles written in TCL start with the string #%Module
```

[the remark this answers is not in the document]
vsoch: I don't think I fully understand then - could you show me before and after for each (with the PR)? 

And regardless, I'd like to refactor the template generation function in the ways I described because there is a lot of redundancy.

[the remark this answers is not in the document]
muffato: Currently, as the documentation says, `default_version` is only implemented for Lmod. It seems its intent is to force people to specify which version they want to load (when set to `false`), and to allow `module load` to arbitrarily choose one version (when set to `true`).

If I've got two `samtools` installed, without my change, `module load samtools` will arbitrarily pick up one of the versions, like Lmod does when the `.version` file is there (i.e. `default_version` set to `true).

What's confusing is that the Lmod implementation needs a `.version` file when `default_version` is `true`, and TCL needs a `.version` file when `default_version` is `false`.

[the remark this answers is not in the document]
vsoch: > Currently, as the documentation says, default_version is only implemented for Lmod. It seems its intent is to force people to specify which version they want to load (when set to false), and to allow module load to arbitrarily choose one version (when set to true).

Yes I think that's accurate - when default version is false the user has to be explicit, and when true we get lmod to choose for us.

> What's confusing is that the Lmod implementation needs an empty .version file when default_version is true, and nothing when default_version is false. The TCL needs a non-empty .version file when default_version is false, and nothing when default_version is true.

I don't think that is true for the current implementation, but the changes that are in these PRs no? For the current, we just generate .version for lmod and ignore tcl. If this is the only way to make it work, I don't see why the implementation details matter - the user just turns on default versions or not.

So could you:

1. Show me the before/after for running the command with tcl (this is your setup, correct?) I want to see how it responds to `please_specify_a_version_number` because if it gets angry at you, I want shpc to set a reasonable default to be the last installed (and allow the installer to tweak as necessary, or enable/disable based on a setting).
2. Please refactor the function to write the version file to not be redundant.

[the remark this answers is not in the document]
vsoch: Give me a few minutes, I'll see if I can help you with a few of the changes I'm asking for. I think it might be more clear if I can show you, and I appreciate the work you've already put into this.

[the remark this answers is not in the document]
vsoch: okay please see https://github.com/muffato/singularity-hpc/pull/1! Ping @marcodelapierre if you want to take a look too.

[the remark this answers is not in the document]
muffato: > I don't think that is true for the current implementation, but the changes that are in these PRs no? For the current, we just generate .version for lmod and ignore tcl. If this is the only way to make it work, I don't see why the implementation details matter - the user just turns on default versions or not.

No, that was right. shpc's `default_version` setting is currently irrelevant for TCL, because `default_version==false` is ignored. It's like `default_version` is considered to be true regardless of what the user wants. In this PR, I make TCL refuse to load a software without its version number if there are multiple versions available and `default_version` is set to false.
 
> 1. Show me the before/after for running the command with tcl (this is your setup, correct?) I want to see how it responds to `please_specify_a_version_number` because if it gets angry at you, I want shpc to set a reasonable default to be the last installed (and allow the installer to tweak as necessary, or enable/disable based on a setting).

I've shown how it works in https://github.com/singularityhub/singularity-hpc/pull/517/files#r830613035 . Without the `.version` file, TCL chooses a version and loads it. With a `.version` file, it tries to load the particular version that is in, and complains if the version is not found. This PR assumes that no version would ever be named 'please_specify_a_version_number`.  I don't understand what else I should be showing.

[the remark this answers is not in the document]
vsoch: @muffato you are responding to the comments / requests I made before I created a PR to your branch. The comments here are now deprecated in response to that.

shpc/main/modules/__init__.py
```diff
@@ -409,43 +396,38 @@ def write_version_file(self, uri, tag, default_choice=False):
 
         # First or last installed
         elif self.settings.default_version == "last_installed":
-            self._default_version_last_installed(version_file, tag)
+            self._set_default_version(version_file, tag)
 
         elif self.settings.default_version == "first_installed":
-            self._default_version_first_installed(version_file, tag, default_choice)
+            first_installed = True if len(os.listdir(version_dir)) == 1 else False
```
vsoch, line 401: This is going to be wrong given an empty directory with just .version. Also given we do this more than once, should we make a common helper function?

shpc/main/modules/__init__.py
```diff
@@ -409,43 +396,38 @@ def write_version_file(self, uri, tag, default_choice=False):
...
         found = [x for x in os.listdir(version_dir) if x != ".version"]
 
         # No versions left, remove the whole directory
         if len(found) == 0:
             self._cleanup(version_dir)
             return
 
-        uri = os.path.basename(version_dir)
+        # With the following settings, .version doesn't contain an actual version
+        # so there is nothing to update
```
vsoch, line 418: This isn’t right - if we are updating an install with default version true we need to change it to the last installed version.

shpc/main/modules/__init__.py
```diff
@@ -409,43 +396,38 @@ def write_version_file(self, uri, tag, default_choice=False):
...
-
-        # Use filesystem dates to derive a tag for the last installed
-        dates = [[utils.creation_date(os.path.join(version_dir, x)), x] for x in found]
-
-        # Sort based on wanting last installed (later time) or first (earliest)
-        reverse = False if self.settings.default_version == "first_installed" else True
-        # reverse True, latest -> earliest (return latest) otherwise return earliest
-        dates = sorted(dates, key=itemgetter(0), reverse=reverse)
-        tag = dates[0][1]
+            tag = found[0]
```
vsoch, line 425: This needs to be sorted depending on the date.

vsoch (commented):
The logic seems off here, but I’m tired so can take another look later.

shpc/main/modules/__init__.py
```diff
@@ -426,6 +426,12 @@ def update_version_file(self, version_dir):
             return
 
         found = [x for x in os.listdir(version_dir) if x != ".version"]
+
+        # No versions left, remove the whole directory
+        if len(found) == 0:
```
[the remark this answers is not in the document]
muffato: If the `default_version` setting is such that a `.version` file needs to be created, it won't be cleaned up by the `_cleanup` function, which only goes up as long as the directories are empty. So we'll end up with a directory that contains no modules, but a `.version`. That's what this `if len(found) == 0` will capture, and then call `_cleanup` again, but this time on the directory that contains `.version`

shpc/main/modules/__init__.py
muffato: Absolutely, and that's another reason why I've introduced the other `_cleanup` above. Note that I didn't touch the logic `True if len(os.listdir(version_dir)) == 1 else False`, just moved it from another function to here, so the weakness you're describing was already there.

shpc/main/modules/__init__.py
muffato: `default_version=true` means we let Lmod/TCL decide what the default version should be. For Lmod that's done by having a `.version` file that is empty or refers to a version that doesn't exist - so no need to update. For TC:, that's done by _not_ having a `.version` file at all

shpc/main/modules/__init__.py
muffato: `found[0]` is used when `len(found) == 1`, see above. That's the same shortcut as there used to be line 439

shpc/main/modules/__init__.py
muffato: (continued)
So: in theory if the directory exists and has exactly 1 entry, it should be an actual software version. But to be safe I can add a helper function if you prefer ?

shpc/main/modules/__init__.py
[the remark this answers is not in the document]
vsoch: So what I'm suggesting is that the cleanup function handles removing the directory and checking for the version file. So the function os.listdir here https://github.com/singularityhub/singularity-hpc/blob/6828a22eb7d821e3c8b5badd5b551766be980918/shpc/main/modules/base.py#L71 instead I think we should have a helper function that checks for "empty" - where "empty" is the logic that we've used several times to list, and not include the version. Then you will just hit the function here, the module dir won't exist, and continue on.

shpc/main/modules/__init__.py
vsoch: My suggestion is to fix the cleanup function to handle the .version file (and nothing else) and then visit this again.

shpc/main/modules/__init__.py
vsoch: Right - but for the design of this main class it should be calling the appropriate function and not making assumptions or decisions early about the module software (that was the idea of the design having these sub-functions that can optionally be implemented by each of tcl and lmod.

shpc/main/modules/__init__.py
vsoch: Let's revisit this after the changes above.

shpc/main/modules/__init__.py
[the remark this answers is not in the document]
muffato: OK, I'm with you :) I can do that

shpc/main/modules/__init__.py
muffato: Sure. I can perhaps make this function more similar to `write_version_file` by adding the series of `if self.settings.default_version ==`. I'll give it a go and will let you know

muffato:
Random thought after seeing https://github.com/singularityhub/singularity-hpc/issues/526#issuecomment-1079399905
The option to select Lmod/TCL for the modules is called `module_sys`. Therefore, would it not be more consistent to define the value of `default_version` that means "fallback to Lmod/TCL" as`module_sys` too, rather than `sys_module` ?

vsoch:
Haha yes, if I used sys_module I definitely meant the inverse to match the settings parameter! And for that particular function name I have mixed feelings - I would say to keep default_version because (intuitively) a function named “module_sys” on a module client should give me the name of the module system, and I’d be surprised to see something running and changing a default version file.

shpc/utils/fileio.py
```diff
@@ -14,6 +14,20 @@
...
+    """
+    if os.path.exists(path):
+        entries = os.listdir(path)
+        if not entries:
+            return True
+        if entries == [".version"]:
+            return True
+    return False
+
+
```
vsoch, line 30: If we want this in utils, it probably should be general (and not customized to one specific use case). E.g.,:
```suggestion
def can_be_deleted(path, ignore_files=None):
    """
    A path can be deleted if it contains no entries, *or*
    if the only files are in ignore_files
    """
    ignore_files = ignore_files or []
    if os.path.exists(path):
        entries = os.listdir(path)
        if not entries:
            return True
        if entries == ignore_files:
            return True
    return False
```
And probably we'd either want to sort both lists, or do an iterator `[x for x in entries if x not in ignore_files]` and then return False if we find any remaining files in that set.

shpc/main/modules/__init__.py
```diff
@@ -58,7 +58,7 @@ def _cleanup(self, module_dir):
 
         # If directories above it are empty, remove
         while module_dir != self.settings.module_base:
-            if not os.path.exists(module_dir) or os.listdir(module_dir):
+            if not utils.can_be_deleted(module_dir):
```
vsoch, line 61:
```suggestion
            if not utils.can_be_deleted(module_dir, [".version"]):
```

muffato:
Ready for review. There is now a single function that writes the `.version` (if needed) according to the `default_version` setting. This implementation lost the optimisation that when doing `shpc install` we know which version is the latest. I can reintroduce that if necessary

vsoch:
Awesome! See my comments above (submit a little bit before)

shpc/main/modules/__init__.py
```diff
@@ -359,3 +363,45 @@ def install(self, name, tag=None, **kwargs):
...
+        utils.write_file(version_file, template.render(version=tag))
+
+    def write_version_file(self, uri):
+        """
+        Create the .version file, if there is a template for it.
+        """
+        version_dir = os.path.join(self.settings.module_base, uri)
+        version_file = os.path.join(version_dir, ".version")
+
+        # No default versions
```
vsoch, line 388: The reason we want to include tag is:

1. to be explicit
2. so all the version setting functions have a consistent set of arguments
3. we cannot be sure how a module system (or future module system) might want to use the tag. It just happens to be the case that we don't need it for specific cases now, but we cannot guarantee this forever.

So I think my preference would be to restore the tag variable, and that way we can theoretically be very explicit about the version (tag) we are providing. I should arguably be able to write my own little wrapper to do some weird custom logic given a specific tag.

shpc/main/modules/lmod.py
```diff
@@ -12,3 +12,13 @@ def __init__(self, **kwargs):
         """
         super(Client, self).__init__(**kwargs)
         self.module_extension = "lua"
+
+    def _sys_module_default_version(self, version_file):
```
vsoch, line 16: module_sys here too!

vsoch (commented):
adding some more notes!

shpc/main/modules/__init__.py
```diff
@@ -359,3 +363,45 @@ def install(self, name, tag=None, **kwargs):
...
+        if self.settings.default_version in [False, None]:
+            self._no_default_version(version_file)
+
+        # allow the module software to control versions
+        elif self.settings.default_version in [True, "module_sys"]:
+            self._sys_module_default_version(version_file)
```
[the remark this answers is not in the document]
muffato: I thought that you were unsure about renaming this one, cf https://github.com/singularityhub/singularity-hpc/pull/517#issuecomment-1079426541 . That's why I left it as it was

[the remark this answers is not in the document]
vsoch: Oh my bad! So I was thinking you were saying to rename the function entirely to just `_module_sys(...)` but in the context here (with the additional `_module_sys_default_version` I'm +1. Sorry for not reading your suggestion carefully enough!

shpc/main/modules/lmod.py
```diff
@@ -12,3 +12,13 @@ def __init__(self, **kwargs):
         """
         super(Client, self).__init__(**kwargs)
         self.module_extension = "lua"
+
+    def _module_sys_default_version(self, version_file, tag):
```
vsoch, line 16: We can set default as None here since it's not required.
```suggestion
    def _module_sys_default_version(self, version_file, tag=None):
```

shpc/main/modules/tcl.py
```diff
@@ -12,3 +13,13 @@ def __init__(self, **kwargs):
         """
         super(Client, self).__init__(**kwargs)
         self.module_extension = "tcl"
+
+    def _no_default_version(self, version_file, tag):
```
vsoch, line 17:
```suggestion
    def _no_default_version(self, version_file, tag=None):
```

vsoch (commented):
This is looking great - I think we are close! I'm still working but I'll try to make some time later for a more thorough review/test, and since we are getting closed to finished I think I'll also ping @marcodelapierre again to see what he thinks!

shpc/main/modules/__init__.py
```diff
@@ -359,3 +363,45 @@ def install(self, name, tag=None, **kwargs):
...
+            # The versions we actually have
+            found = [x for x in os.listdir(version_dir) if x != ".version"]
+            if len(found) == 1:
+                tag = found[0]
+            else:
+                selector = min if self.settings.default_version == "first_installed" else max
+                tag = selector(found, key=lambda x: utils.creation_date(os.path.join(version_dir, x)))
+
+            # Write the .version file
+            self._set_default_version(version_file, tag)
```
vsoch, line 407: Just a suggestion to warrant the extra else block not needed!
```suggestion
        if self.settings.default_version in [False, None]:
            return self._no_default_version(version_file, latest_tag_installed)

        # allow the module software to control versions
        elif self.settings.default_version in [True, "module_sys"]:
            return self._module_sys_default_version(version_file, latest_tag_installed)

        # First or last installed
        # The versions we actually have
        found = [x for x in os.listdir(version_dir) if x != ".version"]
        if len(found) == 1:
            tag = found[0]
        else:
            selector = min if self.settings.default_version == "first_installed" else max
            tag = selector(found, key=lambda x: utils.creation_date(os.path.join(version_dir, x)))

        # Write the .version file
        return self._set_default_version(version_file, tag)
```
Technically they all return None, so the return at the bottom is to be consistent, and the others are to exit early. And maybe one day we will have reason to know some result (e.g., True for changed False otherwise) but we don't see that yet!

I've also walked through the logic of different scenarios, and I don't see any holes! E.g., if we are installing the first time, the module_dir will always create that first version, and then the tag is provided OR we derive at found[0]. There can never be a case where len(found) == 0 because the cleanup function should handle removing empty directories, and the check in uninstall doesn't call the function if the module directory doesn't exist.

vsoch: Also I like the selector bit you added and using min and max,  it's pretty neat :)

shpc/utils/fileio.py
```diff
@@ -5,15 +5,41 @@
...
+    """
+    A path can be deleted if it contains no entries, *or*
+    if the only files are in ignore_files
+    """
+    ignore_files = ignore_files or []
+    if os.path.exists(path):
+        entries = os.listdir(path)
+        if not entries:
+            return True
+        if set(ignore_files).issuperset(entries):
```
vsoch, line 27: Oh I really like this too!

vsoch (commented):
Okay I think this is my final suggestion and then we are ready to merge! We will want @marcodelapierre approval too.

muffato:
I have merged {{main}} into this branch, solved the conflicts, and tested the changes (Lmod and TCL) on my HPC

shpc/main/modules/base.py
```diff
@@ -410,3 +415,44 @@ def install(self, name, tag=None, **kwargs):
...
+            return self._module_sys_default_version(version_file, latest_tag_installed)
+
+        # First or last installed
+        # The versions we actually have
+        found = [x for x in os.listdir(version_dir) if x != ".version"]
+        if len(found) == 1:
+            tag = found[0]
+        else:
+            selector = min if self.settings.default_version == "first_installed" else max
+            tag = selector(found, key=lambda x: utils.creation_date(os.path.join(version_dir, x)))
```
muffato, line 455: There is a potential optimisation here: if `self.settings.default_version == "last_installed"` and `latest_tag_installed` is given, we can directly use it, bypassing the need to list the directory content and fetch their creation date

vsoch: I almost mentioned this in my review! I'm happy if you want to add it, but okay if not.

vsoch (approved):

vsoch (approved):
Okay back to a computer! So I think this is a huge improvement on the current shpc, it adds greater/improved functionality without breaking current setups (since false/true are still supported) and we've had enough back and forth that it's ready for merge, and we can address future issues that might come up! Really fantastic work and work ethnic @muffato I am very impressed (and please know this does not happen too often)! I'll wait until we finish up with the symlink trees to announce these new changes and give you some props on Twitter, etc. <3

marcodelapierre:
Only reading today :)
From what I can tell it is indeed a great addition to SHPC.
Thanks @muffato !

muffato:
Too much praise :)

vsoch:
Nah you deserve it @muffato. We should give others praise more often!

PR #520, opened 2022-03-17
refactor of "add" to generate a container.yaml first to close #519

### LINKED ISSUE #519: Registry Entry that Points to a Local SIF File
I've got a few SIFs that I want to make available to our cluster users via shpc. However, these SIFs are locally stored only (for the time being, hopefully we'll have a gitlab docker registry up and running soon!). 

I'm reading the registry entry writing documentation and I'd like to be able to define aliases, GPU resources, etc for these local sifs as you could in a registry recipe for a remote image. Is that possible?

I've messed with `shpc add` and it seems like it creates just a default setup / module file without the finer control that you get from a registry recipe, is that correct?

vsoch:


vsoch:
@georgiastuart and @marcodelapierre will you have time to review this? I think it's mostly ready to go!

vsoch:
@georgiastuart good to go?

georgiastuart:
We're getting thrashed by storms (Austin, TX) so I unplugged my computer 😂 I'll run through the steps on our cluster ASAP once I'm back online.

vsoch:
Ohno!! Please stay safe!

vsoch:
> Adding a remote registry with a different namespace seems to retain the original namespace when it comes to the modules. For example:

I think I anticipated this - it was a design decision to have the modules derives from the container name, and then to honor the namespace. So should we even support the custom namespace? Do you see reasons, for example, for having two tensorflow installs with the same container? My suggestion / thinking is that we should tweak the docs to not allow a custom namespace for the docker URI and honor/ maintain the namespace until there is a compelling reason to not.

georgiastuart:
If it's a design decision, I think not supporting custom namespaces at all would be better since I would expect it to carry over to the module name. I think it's more confusing to have a registry entry named one thing and the module named another, but that's just my two cents!

So I think the options should either be
1. Support custom namespaces and carry through to module name
2. Don't support custom namespaces at all

vsoch:
Ok great! I totally agree. Let me update the PR to support that.

vsoch:
@georgiastuart all set with changes!

docs/getting_started/user-guide.rst
```diff
@@ -1114,19 +1114,87 @@ Add
...
+    Registry entry vanessa/pokemon:latest was added! Before shpc install, edit:
+    /home/vanessa/Desktop/Code/shpc/registry/vanessa/pokemon/container.yaml
+
+
+And that's it! The container module will use the same namespace, ``vanessa/pokemon`` as the Docker image,
+and we do this purposefully as a design decision. Note that ``add`` previously would add the container directly to the module
+directory, and as of version 0.0.49 it's been updated to generate the container.yaml first.
+Also note that ``add`` is only supported for Singularity, as Docker and Podman containers are 
+typically provided via registries. If you are looking for support for add for another
+container technology, please `open a new issue <https://github.com/singularityhub/singularity-hpc/issues>`_.
```
muffato, line 1196: Hi @vsoch . This seems to contradict the lines above 🤔 . Is Docker supported ?

vsoch: The old add didn't support docker/podman - e.g., the old add would say "take this .sif on my filesystem and manually move it into shpc." The new add (which generates a container.yaml for you) indeed has support!

PR #523, opened 2022-03-22
Update documentation with minor fixes

georgiastuart:
I noticed two minor problems when following the documentation:

1. In the config file, `registry` is a list so we need `shpc config add registry:/<DIR>` instead of `shpc config set registry:/<DIR>`
2. Initializing a config file for the user is `shpc config inituser` instead of `shpc config userinit`

I updated both in this pull request.


vsoch:
These are spot on - thank you!

PR #528, opened 2022-03-26
If the container is deleted, the module has to be deleted too. No need to ask the user

muffato:
This is independent from the symlink and default_version business, so putting this as a separate PR.

Since a module needs its container .sif to function, if the user confirms they want to delete the container, the module would become non functional. I think it's redundant (and misleading) to ask the user if they want the module to be deleted too, `shpc uninstall` should just remove it.

### DIFF
--- a/shpc/main/modules/base.py
+++ b/shpc/main/modules/base.py
@@ -106,21 +106,22 @@ class ModuleBase(BaseClient):
         # Podman needs image deletion
         self.container.delete(name)
 
-        if container_dir != module_dir:
-            self._uninstall(container_dir, "$container_base/%s" % name, force)
-            self._uninstall(module_dir, "$module_base/%s" % name, force)
-        else:
-            self._uninstall(module_dir, "$module_base/%s" % name, force)
+        if not force:
+            msg = name + "?"
+            if not utils.confirm_uninstall(msg, force):
+                return
 
-    def _uninstall(self, module_dir, name, force=False):
+        if container_dir != module_dir:
+            self._uninstall(container_dir, "$container_base/%s" % name)
+            self._uninstall(module_dir, "$module_base/%s" % name)
+        else:
+            self._uninstall(module_dir, "$module_base/%s" % name)
+
+    def _uninstall(self, module_dir, name):
         """
         Sub function, so we can pass more than one folder from uninstall
         """
         if os.path.exists(module_dir):
-            if not force:
-                msg = "%s, and all content below it? " % name
-                if not utils.confirm_uninstall(msg, force):
-                    return
             self._cleanup(module_dir)
             logger.info("%s and all subdirectories have been removed." % name)
         else:
--- a/shpc/main/modules/base.py
+++ b/shpc/main/modules/base.py
@@ -117,6 +117,13 @@ class ModuleBase(BaseClient):
         else:
             self._uninstall(module_dir, "$module_base/%s" % name)
 
+        # parent of versioned directory has module .version
+        module_dir = os.path.dirname(module_dir)
+
+        # update the default version file, if other versions still present
+        if os.path.exists(module_dir):
+            self.write_version_file(module_dir)
+
     def _uninstall(self, module_dir, name):
         """
         Sub function, so we can pass more than one folder from uninstall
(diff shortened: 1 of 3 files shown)


shpc/main/modules/base.py
```diff
@@ -106,24 +106,26 @@ def uninstall(self, name, force=False):
         self.container.delete(name)
 
         if container_dir != module_dir:
-            self._uninstall(container_dir, "$container_base/%s" % name, force)
+            force = force or self._uninstall(container_dir, "$container_base/%s" % name, force)
```
vsoch, line 109: Any reason not to just do:
```suggestion
            force = self._uninstall(container_dir, "$container_base/%s" % name, force)
```

shpc/main/modules/base.py
```diff
@@ -106,24 +106,26 @@ def uninstall(self, name, force=False):
...
         """
         if os.path.exists(module_dir):
             if not force:
                 msg = "%s, and all content below it? " % name
                 if not utils.confirm_uninstall(msg, force):
-                    return
+                    return False
             self._cleanup(module_dir)
             logger.info("%s and all subdirectories have been removed." % name)
         else:
```
vsoch, line 126: I'm weary about this case - if the module directory doesn't exist we sort of "force" the uninstall to be True for the container directory no matter what, and I'm not sure that is best. E.g., I can see someone wanting to re-generate the module file, and they decide to delete the module first (but not the container) and choose yes for the first but not the second. 

If the person chooses `--force` then this should allow them to have a non-interactive "yes to all the things" install, otherwise I think the current setup is what I'd want.

shpc/main/modules/base.py
vsoch: Thinking more about this - if you want just one force, why not move the ask (confirm action) outside of those two functions to just ask once?

muffato: > Any reason not to just do:

Yes, that'd work too :)

> Thinking more about this - if you want just one force, why not move the ask (confirm action) outside of those two functions to just ask once?

My reasoning was to let the user choose whether they want to remove:
 - nothing
 - the module only
 - the module and the container

And thinking that once the symlinks would be added, the four options possible scenario would be:
- nothing
- the symlink only
- the symlink and the module
- the symlink, the module, and the container

i.e. not allowing dangling symlinks or modules, as could happen if we ask about each component separately

shpc/main/modules/base.py
muffato: I'm not following you. shpc first deletes the container, and then the module, cf https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/modules/base.py#L109-L110

By the time shpc checks whether the module directory exists or not, it will have already processed the container directory

muffato:
By the way, what is the reason for the early exit in https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/modules/base.py#L64-L67 ?
I prevents the deletion of empty parent directories. I would have rather skipped `shutils.rmtree` if the path doesn't exist, and then let it enter the _while_ loop.

vsoch:
We are being evacuated for fire I probably won't get to this today, unless the fire is contained. Sorry @muffato !

muffato:
Gosh ! Run !!

shpc/main/modules/base.py
vsoch: Yep, so I think this design was done before we had two things to uninstall - e.g., the confirm was placed in uninstall because it was called first, and when I refactored to have an _uninstall called on the two directories I needed to move the confirmation and check of force up one level! So what I think we should do to fix this is to remove the confirmation from _uninstall and put into uninstall so _uninstall is called already having the confirmation, and from the main uninstall() we do both calls of _uninstall() (if relevant) in one swoop after that.

muffato:
Hi @vsoch 👋🏼 . I added a couple of other commits. I found that empty directories were only deleted for modules, not containers, because the `_cleanup` method was hardcoded to work up to `container_base`. Here is a change to make it work with any directory tree (could be used with symlinks too)

vsoch:
@muffato looks like tests failed because of a simple import error.

muffato:
Sorry. I pushed before doing my tests as well. Now fixed and fully tested on my HPC

shpc/utils/fileio.py
```diff
@@ -61,6 +61,28 @@ def mkdir_p(path):
...
+    """
+    Delete the tree under $path and all the parents
+    up to $base_path as long as they are empty
+    """
+    if not os.path.isdir(base_path):
+        logger.exit("Error: %s is not a directory" % base_path)
+    if not path.startswith(base_path):
+        logger.exit("Error: %s is not a parent of %s" % (base_path, path))
+
+    if os.path.exists(path):
```
vsoch, line 74: Any reason to have this delete up here? Won't it happen in the while loop?

muffato: The delete in the for loop is conditioned to the directory being empty, but at this stage `path` is unlikely to be empty. It will likely contain a container, or a module, or a `bin/`, etc

vsoch (approved):

PR #531, opened 2022-03-27
Add empty __init__.py so registry is copied appropriately by pip

### LINKED ISSUE #530: shpc install not finding recipes
**Describe the bug**

When I use `shpc install` to grab a recipe from https://singularityhub.github.io/singularity-hpc/, it returns `not a known recipe in any registry`. 

**To Reproduce**

```
$ shpc install biocontainers/samtools
biocontainers/samtools is not a known recipe in any registry.
```

**Expected behavior**
`biocontainers/samtools` is installed. 

**Version of Singularity and Singularity Registry HPC Client**

Singularity: v3.8.6
SHPC: v0.0.49

georgiastuart:
Closes #530 

# The Problem

When using `pip install singularity-hpc` or `pip install .` in the source code, the `registry` file is not copied appropriately to `site-packages`, which results in, for example,

```
$ shpc install biocontainers/samtools
biocontainers/samtools is not a known recipe in any registry.
```

# The Fix

A simple fix is just adding an empty `__init__.py` to the `registry` directory so setuptools will copy appropriately. I've confirmed this works with `pip install .` on Rocky 8 in a Conda environment, though I can't test pulling from pypi until the PR is accepted.


vsoch:
That works for me! I don't see any reason the empty init would hurt as opposed to another means to add it, so let's go for that and tweak in the future if anyone has issue.

vsoch (approved):

vsoch:
To just leave the note in this issue - the expected install of shpc is just a clone from git (or release) and I haven't put much time/thought into the pip install - I mostly put it on pypi because that's what I do with all my other package and it's good for versions and hashes of files.

PR #538, opened 2022-04-02
Add support for shpc update

vsoch:
This is early work to add the ability for shpc to update container tags. Currently I just am supporting docker, mainly so we can test and figure out the best ideas for the feature. I'll include some examples (the diff) of doing dry-run for various containers in our collection. Notably, the ones that will "break" are latest, so I 'm thinking we will want a "keep list" of tags that won't be subject to the scheme (not implemented yet).


vsoch:
Here is the biocontainers/samtools I tested - it would say the same but add a few versions back.

image

I decided we shouldn't add anything deemed "earlier" than the oldest, so after that:

image

Here is a simpler one - this is what I'd expect based on no longer allowing "latest"

image

But I'm not super happy - I just tried rocker/ml and got an error in the comparison. Basically we are using Python's ability to compare tuples (or lists of things) and we hit an issue when we find different types (e.g., comparing an int to a string). I was originally being more aggressive and flagging anything for removing that raised an error, but I don't think this approach is good - we need to just remove things we are certain we cannot compare (and not the pair). So I was trying a more detailed comparison, e.g, looping through the two and trying to do the comparison on my own:

```python
    def _cmp(self, other):
        if isinstance(other, str):
            other = LooseVersion(other)

        # We can only compare matching types
        shortest = min(len(other.version), len(self.version))

        for i in range(shortest):
            this_version = self.version[i]
            other_version = other.version[i]

            if type(this_version) != type(other_version):
                continue

            if this_version == other_version:
                continue
            elif this_version < other_version:
                return -1
            elif this_version > other_version:
                return 1

        # If we get to the bottom, consider equal
        return 0
```
that at least doesn't error and seems to be slightly more forgiving?

image

I haven't pushed those changes yet - but @muffato @marcodelapierre let me know what you think! I do think given the weirdness of the versions we will need to do something sane and custom on top of LooseVersion as I'm trying here.

vsoch:
Here is an example with the `--filter` for cuda:

image

Here is another example of why we might want tags, perhaps even stored alongside the recipe for update, our previous choice of tags for "redis" included alpine (and the chosen updates do not):

image

Adding the filter:

image

And an example of a filter just to a specific tag:

image

ok I'm going to stop for now because I think I might be over-thinking and/or just creating a monster! :laughing:

vsoch:
Just thought of another idea (will try after dinner) - if we can't compare the tags (aside from equality) bu they have meaning, when I'm originally parsing the string I can try to take them out, put them somewhere else, and then, for example, I could group all the "cuda" or "mpi" containers together (or similar). People do tend to use strings like that in container names so I think there is an idea in there!

Will try after dinner :)

vsoch:
okay just did a bit more manual (with helper script) testing and I'm feeling surprisingly good about this - for cases with some weird "but we should be doing X instead" it was mostly fixable by adding a filter to the recipe. So given starting with some set of commands in a file, e.g.,
```
shpc update consul --dryrun
shpc update golang --dryrun
shpc update nvcr.io/hpc/autodock --dryrun
shpc update mongo --dryrun
shpc update ruby --dryrun
shpc update vault --dryrun
shpc update adminer --dryrun
shpc update biocontainers/bcftools --dryrun
shpc update rocker/shiny --dryrun
shpc update python --dryrun
shpc update tomcat --dryrun
shpc update node --dryrun
shpc update jupyter/tensorflow-notebook --dryrun
shpc update openjdk --dryrun
shpc update spack/ubuntu-bionic --dryrun
shpc update r-base --dryrun
shpc update bids/brainsautoworkup --dryrun
shpc update bids/tracula --dryrun
shpc update kibana --dryrun
shpc update vclab/star --dryrun
shpc update elasticsearch --dryrun
shpc update dinosaur/salad --dryrun
shpc update gradle --dryrun
shpc update php --dryrun
shpc update nginx --dryrun
shpc update mysql --dryrun
shpc update vanessa/salad --dryrun
shpc update couchdb --dryrun
shpc update ghcr.io/autamus/hpctoolkit --dryrun
shpc update ghcr.io/autamus/gromacs --dryrun
shpc update ghcr.io/singularityhub/github-ci --dryrun
```
The last one is oras. Call that `names.txt` and then I could run (with shpc on the path)

```python
# This is run.py
import os
import time

with open('names.txt') as fd:
    lines = fd.readlines()
    
for line in lines:
    line.strip()
    print(line)
    container = line.replace('shpc update', '').replace('--dryrun', '').strip()    
    path = os.path.join("registry", container, "container.yaml")
    os.system(line)
    print()
    os.system("cat %s" % path)
    time.sleep(10)
```
and that gave me enough time to inspect the container yaml and output and look for issues. And strangely, with the new changes it did pretty well! And I just removed the tags (and stored them) I didn't actually wind up doing any comparison with them. But it has me thinking - it could be a really cool / useful little (different) library to be able to define weird variants of a LooseVersion or StrictVersion with custom steps for parsing, filtering, ordering, etc.

muffato:
Honestly, @vsoch, this looks fabulous ! The code is especially neat.

What does "I just am supporting docker" mean ? Are you talking about the docker API which is used on hub.docker.com but I guess quay.io too according to the screenshots ? What about GitLab and GitHub registries: do they use the same API ? (me being curious 😄 )

vsoch:
> What does "I just am supporting docker" mean ? Are you talking about the docker API which is used on hub.docker.com but I guess quay.io too according to the screenshots ? What about GitLab and GitHub registries: do they use the same API ? (me being curious smile )

Oh so it means (more generally) a URI from an OCI registry (oras: or docker in the container.yaml) is supported - anything that can give us tags from crane.io meaning _any_ docker / OCI registry (GitHub packages and quay and docker hub are all supported!) The small set not supported would be the old `singularity-deploy` that was a hack to release a sif to GitHub releases, which I hope nobody uses now that we can push sif via oras. 

As long as GitHub / GitLad are added via oras, they should work too.

vsoch:
If you think this looks good, what I can do is make a branch off of this one, run some new update workflow locally (that would mimic some future GitHub action to do it) and then open a PR, and then we can inspect all the changes more closely.

muffato:
Yep, good idea 👍🏼

vsoch:
okay posted! https://github.com/singularityhub/singularity-hpc/pull/541 I'm going to do dinner but I'll be back after to carefully look over those set of updates.

vsoch:
woo we did it!!

vsoch:
@muffato just to follow up on something (I mentioned somewhere but can't find it now) I did wind up scratching my itch to think of a library to make it easy to put together a custom processing pipeline - it's in early stages but it's super fun so far! https://vsoch.github.io/pipelib/getting_started/user-guide.html#a-real-world-example-docker-tags

And I realize there is a deprecating warning for distutils.version so I'm going to test out the suggested packaging module next.

PR #541, opened 2022-04-05
preview of changes done by shpc update

vsoch:
I'm opening this against #538 so we only see changes to the container.yaml files! 

I'll look this carefully over tonight and make comments about any questions or issues that I see - just decided I was very hungry and dinner sounds good :)


vsoch:
okay already have a bug! The previous version of python shouldn't have been removed (causing the error here) I'll debug after dinner.

registry/golang/container.yaml
```diff
@@ -3,16 +3,17 @@ url: https://hub.docker.com/r/_/golang
...
-  1.17.3: sha256:199102125d11c943c927a8a33911ef960ca72c4879e307c7c2e40ceaa72201e3
-  1.18-rc: sha256:c9be2380bfdc0f4da95e1635e914dbd596bc73b618cdf16894a631eb58945156
-  latest: sha256:e06c83493ef6d69c95018da90f2887bf337470db074d3c648b8b648d8e3c441e
+  1.16.5-alpine: sha256:45f32e963bb3cc408cfcd01a8e76b2872fb238f602ec5481cd75393da29369c0
+  latest: sha256:5eb58ca0a747ed2e2f4e069d1116badb02a172cf160d31f801776a2342c12863
+  '1': sha256:5eb58ca0a747ed2e2f4e069d1116badb02a172cf160d31f801776a2342c12863
+  1-alpine: sha256:3afd220509acf9866e91932a3a41bf341b8bada82107ef3ecce3422826b98064
+  1-bullseye: sha256:da371518962826141d504029f0e8c27c64f57c69d2899ebf3c9a9e7afb136a10
+  1-buster: sha256:284da7e6ad7fd1250a785c24912fb6f8652176b588d637de5c5b3d4a4aa1693a
+  1-cross: sha256:bb3578192ab4a6b744819de9811d0b494a7b8444490e4c6f13926aa201163925
```
muffato, line 16: That's an interesting one, and sort of the same issue as `openjdk`, `php`, `mysql`, `postgres`, etc. Since `shpc` focuses on reproducibility, ideally it should favour the full versions such as `1.18.0-buster` rather than `1-buster`

And since it only considers the latest _n_ tags, you could add a filter to only list say  the `*-buster` tags. I imagine that on an HPC someone would rather pin the underlying distribution, and get several versions on it, rather than essentially pinning the version and getting all the distro flavours. But I see that'd be a controversial change.

registry/python/container.yaml
```diff
@@ -3,17 +3,10 @@ url: https://hub.docker.com/_/python
...
-  3.9.2-alpine: sha256:f046c06388c0721961fe5c9b6184d2f8aeb7eb01b39601babab06cfd975dae01
-  3.9.2-slim: sha256:ce367bb30b8928efb632c369e3bd4a8dbd7905417bd0a245087a82c250e54a24
-  3.9.4-alpine: sha256:419a7502c95b49946fbdd8228b32243597d9e9f191ddfe5468a3b9b3fe64051d
-  3.9.5-alpine: sha256:7dd8962ad2a63403428d652a64d814a5002f1386379355edf5970e40557fe4e6
-  3.9.6-alpine: sha256:954ea8d05e9041d1dd17b69eb13708a60ef8b8bcc76d928beb4d137e2a9ceb30
-  3.9.7: sha256:8771691756bbf5beff80d64fca8f5b12e018352ddd9e30d8cdfef8cc3717b0e6
-  3.9.9: sha256:dd8335df6162579adadd56ff2b9fbd61199da5405c8856c6e34356c13b48cce4
-  3.9.10: sha256:5ca3e9ca37dce5ac4614bc307e3bf6759cb5f0a4285cad50e05cacb81c9318a5
   3.10.0a7-alpine: sha256:9b7958e47cd5bd4d092c3b28802493ad1870bce988b2f6ff97f6c81d96fcda80
+  3.9.10: sha256:3aae21920963df3205fba69826cc07fcf2fad91f9e062add921766b36e89e6e8
```
muffato, line 9: I see you have the `3.9.*` filter. Do you think it'd be possible to configure the updated to get the latest version of each 3.x ?

registry/quay.io/biocontainers/diamond/container.yaml
```diff
@@ -5,9 +5,8 @@ description: Accelerated blast compatible local sequence aligner
 latest:
   2.0.14--hdcc8f71_0: sha256:275f3b3c587f8a40a39693db8acd91da8be6f053ff1426863da22061bd4e7957
 tags:
-  0.8.26--h2e03b76_4: sha256:1bdf8de97dfdfb795a14cecf22cc7b249801ac9886e187533c3b1cf3c9cb7613
-  2.0.11--hdcc8f71_0: sha256:cb3a52f7f7d745e3c023b40faebd5099afceddb290bd65ebd892ddabef41463a
-  2.0.13--hdcc8f71_0: sha256:2c70b7e6e92116163822053400f8528bae95b8846073fa61224cae14cf1ed0dc
   2.0.14--hdcc8f71_0: sha256:275f3b3c587f8a40a39693db8acd91da8be6f053ff1426863da22061bd4e7957
+  0.9.36--h56fc30b_0: sha256:0f80076288d495263598b24d8684da078a86f359bd58e39958cb491abd817399
+  0.8.36--h8b12597_4: sha256:13476bcb2a7077eae93db47fcc9f7f7662fdcb06193dc95139c7a6cf7c2de824
```
muffato, line 10: It looks funny but that's normal. They skipped v1.* 🤷🏼 
But why are they no other 2.0.* versions ? For a given x.y stem does it only select the highest (z,t) – assuming the pattern x.y.z--*_t ?

muffato (commented):
Some comments. Overall it looks quite nice ! It feels good to see the updates coming through,.

By the way, for all software that are available in both https://hub.docker.com/r/biocontainers/ and https://quay.io/repository/biocontainers/ I would remove the former from the registry. As far as I could see, only the latter is regularly updated from BioConda

registry/golang/container.yaml
vsoch: The reason is because the way we do a comparison (and the way these StrictVersion and LooseVersions work) is by way of tuple comparison. So anything that is a list of numbers we can compare! For example using some of the above:

```
In [1]: [1] < [1,1,1]
Out[1]: True

In [2]: [1] < [1,1]
Out[2]: True

In [3]: [1] < [0,1]
Out[3]: False
```
And so we are technically reading all of those as just a version 1. Here is what I'll try for the next run - doing the same "don't allow duplicates" but not just for major minor, for just major too. So here there is just one version of a 1-<some string> container.

image

This is probably a good idea - I'll have this in for the next run (it sounds like this is a common issue, and akin to have a gazillion versions of 1.17 we should probably not have a gazillion of 1).

registry/python/container.yaml
vsoch: yep will change to `3[.]*`I was trying to pin to just 3 so I would weed out most of the namespace.

registry/quay.io/biocontainers/diamond/container.yaml
vsoch: This is what I'm going to call "too many tag" syndrome. Because crane only gives us the top 50, we only see a sample and not any with 2.0. This is an example that would need to have them manually added. https://crane.ggcr.dev/ls/quay.io/biocontainers/diamond

vsoch:
Bless all the developers that use semantic versions! lol

registry/biocontainers/picard/container.yaml
```diff
@@ -1,12 +1,17 @@
 docker: biocontainers/picard
 latest:
- "v1.139_cv3": "sha256:05a6dd401d47c930191fa8c017b2cf60c099c4da236f134586a2f3610e426229"
+  v2.3.0_cv3: sha256:8624a229b70fc495e29f55f67ec630e30fa43db3f2a5e276e0e0a68551fee0ea
 tags:
- "v1.139_cv3": "sha256:05a6dd401d47c930191fa8c017b2cf60c099c4da236f134586a2f3610e426229"
+  v1.139_cv3: sha256:05a6dd401d47c930191fa8c017b2cf60c099c4da236f134586a2f3610e426229
+  2.3.0: sha256:5e1bc7788a3aa1329e59821e3f89b5bb209372926a67ed1a95f7b9a71350e2be
+  v2.3.0_cv3: sha256:8624a229b70fc495e29f55f67ec630e30fa43db3f2a5e276e0e0a68551fee0ea
+  '1.141': sha256:675128282859c9edf1c8679478c828a48d980a4e73b95abba6a087d315d01cfc
```
muffato, line 8: I'm curious about this. Without the quotes, the yaml parser would transform this to a float ?

registry/elasticsearch/container.yaml
```diff
@@ -16,6 +16,11 @@ tags:
   7.16.2: sha256:fe7ae76ec33e1222ece5216e3a9c6346999a73d3fb65256abb01638758db4b5d
   7.16.3: sha256:de5474d051409e8222eee30c82ef299f3f3c97bb339bbd19c314178db659f7b4
   8.0.0: sha256:19f79b36495507ec53867ab40acaae448a9baa7820fb843be8cb5ede670bac5b
+  8.1.2: sha256:9785077d3781c0049a7fb9488a3b3a6d40bf529c105c8f3530533701ce00e602
+  8.1.1: sha256:dc6cd3eb7d2dcda66f4a1e8535794cda8a201fd80ab16940ab48349356620a19
```
muffato, line 20: I thought that in this case it would only keep 8.1.2 ?

registry/biocontainers/picard/container.yaml
vsoch: All the tags should be parsed as strings and the pieces are split into int.

registry/ghcr.io/autamus/gawk/container.yaml
```diff
@@ -5,10 +5,11 @@ description: gawk is the GNU implementation of awk. The awk utility interprets a
   programming language that makes it possible to handle simple data-reformatting jobs
   with just a few lines of code. https://www.gnu.org/software/gawk/
 latest:
-  5.1.1: sha256:74c9d77c559f704ca0291fc2f98570efbf5b5183bd88cec9c27e51818b71db68
+  5.10.0: 'crane digest ghcr.io/autamus/gawk:5.10.0: MANIFEST_UNKNOWN: manifest unknown'
```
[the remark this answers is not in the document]
vsoch: This will be fixed when I update this PR with our recent fix!

registry/gradle/container.yaml
```diff
@@ -14,7 +14,12 @@ tags:
...
   7.3.3: sha256:8d037a3bd86fc48b591ad778d6eb6130e641bad1be546923e6d097a7ad7708eb
-  7.4.0: sha256:94d9d9e1682b787a5819d5309d04b560b5d9fbd3ba8f94177192a91cc6c22ae6
-  latest: sha256:94d9d9e1682b787a5819d5309d04b560b5d9fbd3ba8f94177192a91cc6c22ae6
+  7.4.0: sha256:5248d0e8f7f6ad2095c3a053d5461daa17b02097410f0e9f6397f8f4dedc34bf
+  latest: sha256:3c50ce7b943306ede1ca0d502a41a8861014ee5377352557716fcc9a8f877992
+  '7': sha256:3c50ce7b943306ede1ca0d502a41a8861014ee5377352557716fcc9a8f877992
+  7-jdk17: sha256:3c50ce7b943306ede1ca0d502a41a8861014ee5377352557716fcc9a8f877992
+  7-jdk16: sha256:f174c0dcf9a84b4035f1fcb62f0340ddc69c0b93320e0d35f097d20ce2ca89d5
+  7-jdk11: sha256:d920da5cedc06e09c031b748673f2f99456fe8a5d3f437e211fd3cf0ca859fcc
+  7-jdk-openj9: sha256:acd908af42e1bee2f841eeac031d41317ab3fddcc3ec9d0d4e1cb4b28be24f5f
```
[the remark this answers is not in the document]
vsoch: This might be the same bug as before with allowing the duplicates - let me run again and we will see if it's still an issue.

vsoch:
okay that did something bad because now it looks like the latest isn't accurate.

vsoch:
I'll need to work on this after work, it's going to be more than a quick push.

registry/jupyter/datascience-notebook/container.yaml
```diff
@@ -3,10 +3,15 @@ url: https://hub.docker.com/r/jupyter/datascience-notebook
 maintainer: '@vsoch'
 description: Jupyter Datascience Notebook from https://github.com/jupyter/docker-stacks
 latest:
-  "4.0": sha256:5e5bf78bfa351c0255cea5269a0461afbf6b50b51e923a7436229208ea8487f9
+  b78201340369: sha256:027e159b6780294aa930623f0f7e6c5094972be48d5e15027024a811fdc7e5c7
```
vsoch, line 6: @muffato this looks like a red flag - I'm wondering if we should try to filter out patterns that look like commits.

vsoch:
okay I think this is more sane - I had a bug that was removing most versions :P I think we might still want to try and remove "stuff that looks like a commit hash" in which case we should figure out a pattern for that, some length of lowercase letters and numbers I think?

vsoch:
Don't worry about the failure for now - it's a test that needs to be updated given we change a version.

registry/jupyter/datascience-notebook/container.yaml
muffato: Yes, you're right. This looks like a commit or a checksum. It's not something that `shpc` could rank against the other tags
(especially this particular one is 2 year old https://hub.docker.com/layers/datascience-notebook/jupyter/datascience-notebook/b78201340369/images/sha256-027e159b6780294aa930623f0f7e6c5094972be48d5e15027024a811fdc7e5c7?context=explore whereas other tags are less then a week old. Hopefully they're legacy tags from a time when they didn't have proper versioning in place ?)

muffato: And to answer https://github.com/singularityhub/singularity-hpc/pull/541#issuecomment-1090781733 

Let's maybe allow strings of length <=8 made of digits only because some project may use timestamps e.g. 20220406.

Option 1. To exclude commit IDs: ignore the strings of length > 8 that are only composed of digits + hex characters only.

Option 2. Am I understanding right that you need some integers to be able to rank the versions ? You could do something after the parsing and splitting. Could we say that the tuple must start with at least 1 integer and that the lengths of the integers must be 8 characters at most ? So it would:
- allow '4.b78201340369' because it looks like a version 4 something, which can still be compared against other version numbers
- exclude all commit IDs that include a letter
- include strings that are entirely made of digits as long as they're not too long

registry/biocontainers/picard/container.yaml
muffato: So if the quotes are unnecessary, can they be removed ? I find it disturbing that some tags have quotes and some don't

vsoch: The quotes are added by the yaml parser, and we use a specific one that preserves comments. I don't know any way to remove them, but I don't think they hurt. Is there a reason to not want them there other than aesthetic?

vsoch:
okay just pushed an updated version that does a pretty good job filtering out the commit tags I think.

registry/mongo/container.yaml
```diff
@@ -15,8 +15,13 @@ tags:
...
-  5.0.6: sha256:c7f5b78eead63e672176d4a6e0932f8649c8f727dec5acef6b9ed7c9bdc3dae5
-  latest: sha256:c7f5b78eead63e672176d4a6e0932f8649c8f727dec5acef6b9ed7c9bdc3dae5
+  5.0.6: sha256:fed6248ae0bb0d54c0448eb786c87120737eedc522172ee1536ad47789782348
+  latest: sha256:fed6248ae0bb0d54c0448eb786c87120737eedc522172ee1536ad47789782348
+  '5': sha256:fed6248ae0bb0d54c0448eb786c87120737eedc522172ee1536ad47789782348
+  5-nanoserver-ltsc2022: sha256:28b9079c157d0f2c23fec6352a37d85e2c20947c721214024b3aea6338341ff7
+  5-windowsservercore-ltsc2016: sha256:9620d60541d45e913c30f78623ee4b5ec8f650399339982dda11e9cd1d1b9c78
+  5-nanoserver-1809: sha256:89c4b070fbc207f5b28f69cff5dec9b841ace73c3c7baeb09b598c060f173699
+  '5.0': sha256:fed6248ae0bb0d54c0448eb786c87120737eedc522172ee1536ad47789782348
 aliases:
```
muffato, line 25: Why is it inserting the flavours of `5` but not the flavours of `5.0.6` ? Is it because `nanoserver` etc end up like minor version number ?

muffato:
> The quotes are added by the yaml parser, and we use a specific one that preserves comments. I don't know any way to remove them, but I don't think they hurt. Is there a reason to not want them there other than aesthetic?

OK, I understand that the need to use that parser trumps the quoting style it uses. No worries

registry/nvcr.io/hpc/autodock/container.yaml
```diff
@@ -6,10 +6,11 @@ description: The AutoDock-GPU Suite is a growing collection of methods for compu
   of the basic mechanisms of biomolecular structure and function. More info on AutoDock-GPU
   be located at https://ccsb.scripps.edu/autodock/ and https://github.com/ccsb-scripps/AutoDock-GPU#usage.
 latest:
-  "2020.06": sha256:a968fa5391281f168e1daf5a4fccfbd0d8228d794ee693ba1a637e5147c3b798
+  '2020.06': sha256:a968fa5391281f168e1daf5a4fccfbd0d8228d794ee693ba1a637e5147c3b798
 tags:
-  "2020.06": sha256:a968fa5391281f168e1daf5a4fccfbd0d8228d794ee693ba1a637e5147c3b798
+  '2020.06': sha256:a968fa5391281f168e1daf5a4fccfbd0d8228d794ee693ba1a637e5147c3b798
   2020.06-arm64: sha256:2d881faf0b20866826a2f0ebbd87ed9791b64222489814901a090e5ed1aa3252
+  2020.06-x86_64: sha256:0bc28fe0b15b172241f785079782447251139ad0ef0b978f3f1483d5caa9e56d
```
muffato, line 13: I'm very glad that this one made it through. It shows that the commit-id filter is not overzealous 👍🏼

registry/openjdk/container.yaml
```diff
@@ -4,13 +4,18 @@ maintainer: '@vsoch'
...
   17.0.1: sha256:0da39ed69dec14f9603e2b916592691cc39341510abdf4255abb1c90b00eb3f4
-  17.0.2: sha256:2c3ffd17a95947a187cbf00b454656566063cc53f21ea419929c0d23f3347572
-  latest: sha256:2c3ffd17a95947a187cbf00b454656566063cc53f21ea419929c0d23f3347572
+  17.0.2: sha256:3d88de143bdd54d3e873302355564818e0ac0d17dbacc9869090197b7dd8a8f3
+  latest: sha256:3d88de143bdd54d3e873302355564818e0ac0d17dbacc9869090197b7dd8a8f3
+  '19': sha256:7e593cb75001103d515e7f3a1fc81598c5d7753339368399b7b5226816036c7f
+  19-ea-10: sha256:1dceb32239eb8a2558be7bbc6fcf53b9759799b99a75f1f3a73f483cc79ac028
+  19-alpine3.15: sha256:fd8c856fb7e3fc2d93815c4b9fa1d112c8d9f46d01f64c6ea858d33518555e5f
+  19-ea-15: sha256:7e593cb75001103d515e7f3a1fc81598c5d7753339368399b7b5226816036c7f
+  19-ea-14: sha256:ad5eab9ef6f4e4d0dd971ea0e94eaf992f4d8fb1853ee5d8dfac7e273f4dfed3
```
muffato, line 18: I'm not actively developing with Java, but as far as I can see, `-ea` means "early access". Would it make sense to completely filter them out ? Otherwise, I think there should ideally only be the last one, i.e. `19-ea-15`, but it's probably not worth modifying the shpc code for that

registry/quay.io/biocontainers/trinity/container.yaml
```diff
@@ -5,12 +5,15 @@ description: Trinity, developed at the Broad Institute and the Hebrew University
...
   of transcriptomes from RNA-seq data.
 latest:
-  2.13.2--ha140323_0: sha256:727ae4211b1a2fdfa33c1bb30fbf8ffc231d610c6497fe7150b031fa9a70b7b1
+  date.2011_11_26--ncurses5.9_8: sha256:eabe346a2adac1bb30aaaa62764f299638daa2a83ef6408722b0f89f8c490e20
 tags:
   2.11.0--h5ef6573_1: sha256:6d35d716aa12ba7b0c715fa8a30359b43ba9151e854dc4407b949e7b57c3a50a
   2.12.0--h5ef6573_0: sha256:4045560dad56fee946c9bcccd2ba0e373292a5adf8a2f7ba09917cbaeaa74323
   2.12.0--ha140323_1: sha256:2736617a96c30bce93a149fe5f47b4f4a290a7d21c1503619430db619d9bc3a0
   2.13.2--ha140323_0: sha256:727ae4211b1a2fdfa33c1bb30fbf8ffc231d610c6497fe7150b031fa9a70b7b1
+  date.2011_11_26--ncurses5.9_8: sha256:eabe346a2adac1bb30aaaa62764f299638daa2a83ef6408722b0f89f8c490e20
```
muffato, line 14: This one is 4 years old. Can it be filter out ?

muffato (commented):
Thanks for working on this, @vsoch . A last batch of comments :)

registry/mongo/container.yaml
vsoch: We only consider the major and minor (e.g., 5.0) so if the parser has already seen 5.0 (without a third) it will not include 5.0.6. The idea behind that is that 5.0.6 might be deprecated soon, but the 5.0 will be a "last update of 5.0 with all changes for that version family). This also assumes that the person is not saying that 5.0 == 5.0.0 - if they used 5.0.0 it would be ordered as less than 5.0.6, and that would be chosen first.

registry/openjdk/container.yaml
vsoch: Ah we can add a custom filter here to avoid tags with ea!

registry/quay.io/biocontainers/trinity/container.yaml
vsoch: Definitely.

vsoch:
okay I wound up doing two changes:
 
- remove the 2011 date here https://github.com/singularityhub/singularity-hpc/pull/541/commits/d2746248854f887e7e7fedd2bb5498ef6be179f5
- redid the openjdk because I think we probably don't need to provide windows containers for hpc :) https://github.com/singularityhub/singularity-hpc/pull/541/commits/a0ee49a9962a1ba446bb414fa139cb6a7de6f372

muffato (approved):
I think you got it 👏🏼

vsoch:
Agree! And I'm good to tweak the recipes of individual containers as we move forward.

PR #545, opened 2022-04-16
Add/views

vsoch:
This is the first pass at support for views to shpc! Specifically this means:

- You can create one or more named views that are for subsets of models. Per the definition of a view, we always create symlinks, and always with shortened paths.
- You can create, delete, edit, install, and uninstall

I've written quite a bit of documentation in the user guide so I won't repeat it here. I'll keep updating the PR until the tests are working, etc.

More advanced functionality I can either add here (or for a later PR is):
- [x] creating a view from a view.yaml file
- [x] `shpc view <name> list` to list what is installed in a view
- [x] figuring out support for system_modules (I suspect we will need to not create a symlink but a copy of the module file, edited with the requirements, unless there is another way). [Could this maybe help](https://lmod.readthedocs.io/en/latest/070_standard_modules.html)?


vsoch:
Huh so it looks like the merge into my symlink tree install (after changes) the CI was broken w.r.t default versions - this is the same issue we see there! Debugging now.

vsoch:
ping @marcodelapierre @muffato ! I'm not sure how I did it (intense avocado energy?) but I finished the first implementation of shpc views in one evening! :partying_face:  I wrote some pretty extensive docs:

https://github.com/singularityhub/singularity-hpc/blob/add/views/docs/getting_started/user-guide.rst#views

and probably the best thing to do is start there, test out what I've written, and then come back and review here. Note that I've done some considerable refactoring of the ModuleBase to (instead of doing "all the things" to have its own template and versionfile class, and that way they can be used by the new views module too! 

I tried to implement the bulk of what we discussed, and some things are left out with the intention to figure out adding (e.g., the system_modules is there but we need to find a strategy for the view to always load them alongside, meaning something in addition to the module symlink or perhaps an actual copy of the file with new content added - see a link I posted above in the description for an idea). And some of the other features in the long threads I did not add, intentionally. 

## Things intentionally not added

- tree-naming and "short" vs. "long" install trees - since views are smaller trees and the shorter names are better, I see 80% of cases wanting that and don't see reason to support installing to a long path. We could add this is we find people have compelling need (and ask) for it
- view-type or "install the full thing instead of a symlink" this to me feels like over-engineering, especially with another settings.yml file. If we need to have further customization of the views, it can be in the view.yaml, but approached very conservatively because it is technically a view of the main install. I would argue that anything that is not a variant of the original install (a view) should be created via an `shpc install` directed at a different module_base, and done. Creating a full copy of something doesn't fully honor the idea of a view, based on a database view, that it should be a splice / subset of the install and using the same core.

## Specific Questions/Requests

The following are details that we talked about in some thread and I'll attempt to review/ summarize here, and please bring up for further discussion those that need it!

- I think this will fit the use case I had in mind of being able to generate a view for a workflow manager (or similar), because we can install from a view.yaml file that could be kept with the workflow, which will work if shpc is installed, and given that the running user has permission to write to the module / container directories. I do _not_ require that module to be installed in the module_base for the reason of easier usabiity - if it doesn't exit it can be installed in both at the same time (actual module and symlink). Personally speaking, if I had to do `shpc install` before every `shpc view install` it would be really annoying.
- burden of sysadmins - I think this could be helped if the requesting user simply provided a views.yaml with the views they want, then it's just one command. Of course further management would need to be done by the admin, but it's not super terrible. It could be the case that the admin allows the user to put the view.yaml somewhere, and it's automatically updated from at some interval?
- the view that has "all software" is afforded by defining a default_view, meaning anything you install to shpc automatically is also symlinked there. You do have to handle conflicts if/when they arise, but I don't think that would be super common.
- "disk hungry containers" are handled because there is only one container base to serve all views
- "load" and "unload" aren't really relevant because it's up to the module software (e.g., module use) to decide what is "loaded." shpc view can help with `shpc view <name>` to provide a fullpath for the view for that.
- views do not have their own settings.yaml. There is only one primary "shpc level" settings.yaml that can be modified by a user, and it's limited to that. If we forsee wanting some further customization of a view (in any way) it would be on the level of the views.yaml, and we should pursue this conservatively and carefully.

## Allowing custom settings for views

For this use case:

>    same cluster, multiple GPU vendors -> need a different GPU setting

Here is a simple idea. I suppose we could also have a custom settings portion of the view where you can modify a subset of settings attributes, the caveat being it will ALWAYS take preference, e.g.,:

```yaml
view:
  name: mpi
  modules:
  - ghcr.io/autamus/clingo:5.5.1
  - ghcr.io/autamus/emacs:27.2
  system_modules: []
settings:
  features:
      mpi: nvidia
```
E.g., the above says "I don't care what the user or default settings.yaml says, this view ALWAYS uses mpi with nvidia. But the reason this is "murky" for me is because we can no longer just symlink the module file - we are essentially creating a separate shpc install that may use the same container (with its own module file). I'm not sure that feels in spirit of the view, but I do see the use case. Maybe we would allow customization (and re-generation of a new module file) but not pulling another container, because that is redundant. That will make the implementation quite more complex than just needing to symlink I'm afraid.

vsoch:
ping @marcodelapierre @muffato in case you didn't see this!

marcodelapierre:
I did see the notification @vsoch, and I am very excited about what you have just done! I hope to find the time soon to have a deeper look at this PR! :-)

vsoch:
okay perfecto, that works for me!!

muffato:
I saw it too, and it looks very interesting, but I haven't tried it yet

shpc/client/__init__.py
```diff
@@ -141,6 +172,32 @@ def get_parser():
...
+        description="view control to create, install, and uninstall",
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+    view.add_argument(
+        "params",
+        nargs="*",
+        help="""View control. A view name is always required.
+shpc view create <name>
+shpc view delete <name>
+shpc view get <name>
```
muffato, line 186:
```suggestion
shpc view get <name>
shpc view list <name>
```

muffato:
HI @vsoch,

I've given it a try, and it's very nice to use. It feels like another barrier has been lifted, by allowing as many (symlink) views as needed, for different themes, purposes, etc.

Two bugs I've found:
1. `shpc view uninstall` doesn't seem to update `view.yaml`
2. `shpc uninstall` doesn't check if the module isn't used in any views, so it can break views

Additional questions:
1. Is there an CLI mode to list all existing views ? It looks like the function `detect_views` could do that, but I couldn't find how to trigger it.
2. What is the purpose listing all modules in the`view.yaml` files, and why is it deemed needed for views but not for the complete shpc installation ?

I've only tried adding / removing views, and software in it. I haven't checked that the other features that were discussed, such as `default_version`, still work.

muffato:
`default_version` seems to work too 👏🏼

vsoch:
Nice! Thank you @muffato ! I'll work on those bugs this evening! :bug:

vsoch:
> 1.  Is there an CLI mode to list all existing views ? It looks like the function detect_views could do that, but I couldn't find how to trigger it.

I will add one! I think this occurred to me when I was writing it and I forgot to add.

> 2. What is the purpose listing all modules in theview.yaml files, and why is it deemed needed for views but not for the complete shpc installation ?

Yes! It's because at any point in time you should be able to share that file and someone can install / re-create your view from it.

vsoch:
> shpc uninstall doesn't check if the module isn't used in any views, so it can break views

I wasn't sure what would be best to do here. Should we just ask for confirmation, and give the user a warning about the number of views impacted (and then uninstall from there too?) Or prevent uninstall all together and force the user to uninstall from the view first?

muffato:
> > 2. What is the purpose listing all modules in the `view.yaml` files, and why is it deemed needed for views but not for the complete shpc installation ?
> 
> Yes! It's because at any point in time you should be able to share that file and someone can install / re-create your view from it.

`shpc list` does this without having to maintain a file – by traversing the module tree and parsing the paths. In my view, the difference with views is that views are symlinks, so `shpc` would have to `readlink` all the symlinks, but it could then parse the paths the same way, and get the live list of modules of a view



> > shpc uninstall doesn't check if the module isn't used in any views, so it can break views
> 
> I wasn't sure what would be best to do here. Should we just ask for confirmation, and give the user a warning about the number of views impacted (and then uninstall from there too?) Or prevent uninstall all together and force the user to uninstall from the view first?

👍🏼   on letting the user know which views are affected, and then removing the module from all upon confirmation (which should be skipped in "force" mode ?)

vsoch:
> shpc list does this without having to maintain a file – by traversing the module tree and parsing the paths. In my view, the difference with views is that views are symlinks, so shpc would have to readlink all the symlinks, but it could then parse the paths the same way, and get the live list of modules of a view

I think the expectation is that we want to add more than just a list of modules to the view - e.g., "always load these system modules" or other customization to views. It should always be possibly to quickly grab that to share. It does seem redundant right now, but I think having a config file for each view (that you can just forget about, it updates with shpc) is a good approach. We shouldn't need to rely on parsing the filesystem just for the listing of what is installed, imho.

> 👍🏼 on letting the user know which views are affected, and then removing the module from all upon confirmation (which should be skipped in "force" mode ?)

Gotcha! I'll add this listing of what modules have it to shpc uninstall. It's almost the end of the work day so I should be able to soon!

vsoch:
okay all set @muffato

muffato:
It works for me 🚀

vsoch:
Wicked! Let's give @marcodelapierre a chance to take a look. Importantly, we do need to discuss the "additional features" that are desired for a view. E.g., adding additional system modules is something we might want to be able to do, but we'd need to figure out if there is a way to load them in a different file (akin to the .version file so we can keep the symlinks and don't need to re-create the files in views).

I'm also really happy with how this is coming out - I like this solution better than the one hard coded symlink tree!

muffato:
Can you clarify what you mean by "system modules" ? My understanding so far is that it's a set of modules not managed by shpc, that are dependencies of shpc modules. Is that right ?
In my mind, they're a generalisation of the `singularity_module` and `podman_module` option, but that would be recorded at the view level. For instance, openmpi vs mpich, etc, cf #527

TCL modules accept `module load` directives in them, but also listing required modules (which prevent loading if the dependency is not loaded yet). Which behaviour were you expecting to implement ?

vsoch:
Sure! So an easy example is for a view named "mpi" and then wanting to load whatever system module (e.g., openmpi version X.X) along with it.

> In my mind, they're a generalisation of the singularity_module and podman_module option, but that would be recorded at the view level. For instance, openmpi vs mpich, etc, cf https://github.com/singularityhub/singularity-hpc/issues/527

Yes exactly. The user should be able to either manually (or with shpc view) add an attribute like a system_module. In fact the list is already there but not being used:

```yaml
view:
  name: second-mpi
  modules:
  - ghcr.io/autamus/emacs:27.2
  system_modules: []
```

> TCL modules accept module load directives in them, but also listing required modules (which prevent loading if the dependency is not loaded yet). Which behaviour were you expecting to implement ?

We can't just change the base recipe file because it's a symlink, and that would make the changes for all modules (not just those in the view). That's what I'm not sure about - how we can add additional directives without disturbing the symlinked file. E.g., some file in the root that is always loaded in addition to the symlinked files below it? :thinking:

muffato:
It might be possible to write Lua/TCL code that parses the path of the module, recognises symlinks, checks the parent directories for mandatory modules to load.

That said, would it make sense to have a module added to different views with different system modules ? The main driver seems to be MPI implementations: can a single software image be compatible with multiple MPI implementation ??
(I'm wondering if the system modules should be a property of the module, not the view ...)

vsoch:
I think you could probably see both cases.

marcodelapierre:
Hi guys, finally dedicated some time to read through this and the Views docs, and think through.

It looks amazing! Most of the comments/ideas/hints I had, you have already implemented them.

So, just 2 comments, a short one and a longer one.

The short one, on installing modules from a pre-existing view.yaml.
Would this also perform the upstream installation (container pull and upstream modulefile creation), in the case they were not pre-installed?
This might be a nice feature, basically it would enable maintaining a list of container modules by means of a views.yaml.

And following up to this, when there's a default view, is the corresponding yaml updated at every installation? 


Next comment on the longer topic: system-modules/ mpi/gpu use cases

marcodelapierre:
Let me comment on system_modules and its use cases.
Please feel free to comment point by point, happy to chat on this further.

Based on my experience, I would argue that, by definition and design, containers typically encapsulate all required dependencies within them, so there should almost never be a need for loading extra modules when using them.
There may be cases, though, when modules are used to setup specific environment variables; the most striking case I can think of is MPI (although this might just be embedded in the singularity module itself).

This being said, let me first bring back one feature suggestion that is UNRELATED to Views.
I think I brought it up when we were talking with Georgia on the MPI setup, in another issue.
With the preamble above, I would see a possible `mpi` feature as being controlled within the settings.yaml as indeed a list of modules to load, in substitution for the standard `singularity_module` or relatives:
```
features:
  mpich: [<list of modules to load>] or Null
  openmpi: [<list of modules to load>] or Null
```
These would be injected in the modulefiles once and for all, a bit as for the GPU case (--nv or --rocm flag), only for those containers whose yaml has a matching feature, `mpi: mpich` or `mpi: openmpi` (these 2 are ABI incompatible, so a container would be built with one or the other one).
Although I can't see a user case for this, we could also think of a similar `system_modules` setting, with a list of "other" modules that need to be loaded inside the modulefile.

What I described above would do it when only one MPI type is on a system. (similar as the current GPU functionality, which works for one GPU type only).

Now, what if we need more flexibility in a multi-vendor/multi-provider cluster? (I think this was @georgiastuart  use case).

This is where the Views might help, right?
Let me describe a possible way of doing it while keeping the symlinked modules constraint, so no change in the actual modulefiles.
In this case, mpi/gpu settings in the main settings would be left empty, whereas we would have a feature description in the `views.yaml`. 
In the example of MPI, it  could be the same syntax as above, but in views.yaml (and same story for GPU).

How to implement this?
Alongside the view, we could create a view meta-module, called the same way as the view, which does the following:
- loads all container modules within the view
- for MPI, loads corresponding system modules
- for GPU, exports the variable SINGULARITY_COMMAND_OPTS with the appropriate value
- if we want to support generic system_module loads as hypothesised above, then also loads the corresponding modules

This way, the end user loads the meta-module and gets all of the modules plus the desired extra settings; under the hood, this is still a pure symlinked tree.

What I don't like is that this forces people to load all container modules at once, which may give rise to issues/conflicts between aliases/wrappers.
However, right now the only alternative I can think of is to create a brand new moduletree, not a symlinked one, with the appropriate edits in the modulefiles.

I don't have feedback right now on the usage of `.version`  or similar to trigger extra module loads, I haven't looked into that yet.

Thoughts?

vsoch:
Thank you for your review @marcodelapierre ! Here are some quick answers.

> The short one, on installing modules from a pre-existing view.yaml. Would this also perform the upstream installation (container pull and upstream modulefile creation), in the case they were not pre-installed?

Yes - for right now if you try installing to a view (and it's not upstream) we install to upstream first. I think for the most part the users will be admins and it would be annoying more than anything else to get an error and then have to re-issue two install commands, one to the core then one to the view. If we need to customize this in the future (e.g., someone wants the warning instead) that should be easy to add.

> Based on my experience, I would argue that, by definition and design, containers typically encapsulate all required dependencies within them, so there should almost never be a need for loading extra modules when using them.

Ahhhh except for MPI!! And other core libraries that need to work with a container (that may need a custom module load) but I think you are going to talk about this next :)

> There may be cases, though, when modules are used to setup specific environment variables; the most striking case I can think of is MPI (although this might just be embedded in the singularity module itself).

bingo! Stupid MPI. Just kidding. It's alright :laughing: 

```
features:
  mpich: [<list of modules to load>] or Null
  openmpi: [<list of modules to load>] or Null
```

So the use case for this feature in the settings.yaml is a _global_ definition, e.g., "my entire cluster has this kind of mpi and I always want the module installed like this. I think this use case breaks when, say, you have different kinds that you want to load associated with different modules (and no global setting). I disagree this is unrelated to views - I think this case of needing this custom loading or other customizations for some set of groups of modules was what led me to step back and rethink the symlink idea into this idea of views.

> These would be injected in the modulefiles once and for all, a bit as for the GPU case (--nv or --rocm flag), only for those containers whose yaml has a matching feature, mpi: mpich or mpi: openmpi (these 2 are ABI incompatible, so a container would be built with one or the other one).

yes correct!

> Although I can't see a user case for this, we could also think of a similar system_modules setting, with a list of "other" modules that need to be loaded inside the modulefile.

The use case for this is what I described about, when you have other customizations you want to do for groups of modules. So maybe I even have the same set, but one I load with MPI software A and the second with MPI software B (because users requested it).

> What I described above would do it when only one MPI type is on a system. (similar as the current GPU functionality, which works for one GPU type only).

yes exactly!

> Now, what if we need more flexibility in a multi-vendor/multi-provider cluster? (I think this was @georgiastuart use case).

yes yes yes!

> This is where the Views might help, right?

More yes!!!

> Let me describe a possible way of doing it while keeping the symlinked modules constraint, so no change in the actual modulefiles. In this case, mpi/gpu settings in the main settings would be left empty, whereas we would have a feature description in the views.yaml. In the example of MPI, it could be the same syntax as above, but in views.yaml (and same story for GPU).

That could work, but we need a way to implement it without needing to write a different module file. E.g., right now the preference is sort of "hard coded" into the module file (that is then symlinked with it).

> How to implement this?

Are you reading my mind? :laughing: 

> Alongside the view, we could create a view meta-module, called the same way as the view, which does the following:
>    loads all container modules within the view
>    for MPI, loads corresponding system modules
>    for GPU, exports the variable SINGULARITY_COMMAND_OPTS with the appropriate value
>    if we want to support generic system_module loads as hypothesised above, then also loads the corresponding modules
> This way, the end user loads the meta-module and gets all of the modules plus the desired extra settings; under the hood, this is still a pure symlinked tree.

Ohh, that's really interesting! So how would we still ensure this "meta module" is loaded when a view module is loaded? Isn't this the same problem but just mapping it into another module? Would we require them to load the view + a module in it always?

> What I don't like is that this forces people to load all container modules at once, which may give rise to issues/conflicts between aliases/wrappers. However, right now the only alternative I can think of is to create a brand new moduletree, not a symlinked one, with the appropriate edits in the modulefiles.

Is there no way to add a file a level above a module file (that is symlinked) that is always loaded? What about if we added extra logic to a .version file or similar to load this "meta module" with whatever view preferences we need?

> I don't have feedback right now on the usage of .version or similar to trigger extra module loads, I haven't looked into that yet.

Reading my mind again! :laughing: I'm much more terrible with modules than y'all but I can look into this too - it feels like if we can get the software to look at the .version files we might be able to ask them to do other things too. Pinging @xdelaruelle - we are interested to ask if there is a way to have a module file some level about a symlinked module to load required modules for it (so we don't need to tweak the symlinked file?)

> One final note: I proposed to ALSO add the alternative mpi solution above, without Views, because it's simpler, and I liked the idea of providing a bit of flexibility to sys admins, in being able to setup these things in different ways depending on needs: a simpler setup for simpler scenarios, and a slightly more complicated setup (with Views) for more complicated scenarios.

Oh I think we can still tweak that particular setting - I wanted to do this first because we have quite a bit of PR energy here and it would get harder to merge into main if we abandon it for too long!

xdelaruelle:
> Reading my mind again! laughing I'm much more terrible with modules than y'all but I can look into this too - it feels like if we can get the software to look at the .version files we might be able to ask them to do other things too. Pinging @xdelaruelle - we are interested to ask if there is a way to have a module file some level about a symlinked module to load required modules for it (so we don't need to tweak the symlinked file?)

You may use the [ModulesCurrentModulefile](https://modules.readthedocs.io/en/latest/modulefile.html#mfvar-ModulesCurrentModulefile) variable within the modulefile to make it aware of its own fullpath name. Then adapt the evaluation, to define specific requirements based on this value.

Not sure I correctly get what you are trying to do. If you have somewhere a comment or notes on what you are trying to achieve at the module level, I would be happy to read that.

marcodelapierre:
Thanks @vsoch I think I agree along your whole line.

To answer this point:
> Ohh, that's really interesting! So how would we still ensure this "meta module" is loaded when a view module is loaded? Isn't this the same problem but just mapping it into another module? Would we require them to load the view + a module in it always?

My suggested solution unfortunately doesn't solve this one directly.
The users would need to load the whole metamodule, eg `module load mpi` (with `mpi` being the name of the example view), to get the whole view modules loaded, plus the additional customisations.

That's another reason why it would be great if we could achieve this with some other module config trick, which leaves the user experience unchanged, ie the usual load package specific modules as needed.

vsoch:
oh right, sorry let me provide a little more detail! So right now shpc serves to install container modules - meaning a module file with somewhat hard coded interactions with the containers. To the user, they just load a variant of python and run python, but (thanks to the magic of the modulefile) they are actually running a container exec command, as an example. So this workflow looks like:

```
shpc install <container module> -> container.sif plus module file --> module load --> python!
```
And the filesystem layout might look like:

```bash
modules/
├── dinosaur
│   └── salad
│       └── latest
│           ├── 99-shpc.sh
│           └── module.lua
├── ghcr.io
│   └── autamus
│       ├── emacs
│       │   └── 27.2
│       │       ├── 99-shpc.sh
│       │       └── module.lua
│       ├── rust
│       │   └── 1.54.0
│       │       ├── 99-shpc.sh
│       │       └── module.lua
│       └── samtools
│           └── 1.14
│               ├── 99-shpc.sh
│               ├── bin
│               │   ├── blast2sam.pl
│               │   ├── bowtie2sam.pl
│               │   ├── export2sam.pl
│               │   ├── fasta-sanitize.pl
│               │   ├── interpolate_sam.pl
│               │   ├── novo2sam.pl
│               │   ├── psl2sam.pl
│               │   ├── sam2vcf.pl
│               │   ├── samtools
│               │   ├── samtools.pl
│               │   ├── seq_cache_populate.pl
│               │   ├── soap2sam.pl
│               │   ├── wgsim_eval.pl
│               │   └── zoom2sam.pl
│               └── module.lua
├── quay.io
│   └── biocontainers
│       └── trimmomatic
│           └── 0.39--hdfd78af_2
│               ├── 99-shpc.sh
│               └── module.lua
├── redis
│   └── 7.0-rc
│       ├── 99-shpc.sh
│       └── module.lua
├── ruby
│   └── 3.1.1
│       ├── 99-shpc.sh
│       └── module.lua
└── vanessa
    ├── pokemon
    │   └── latest
    │       ├── 99-shpc.sh
    │       └── module.lua
    └── salad
        └── latest
            ├── 99-shpc.sh
            └── module.lua
```
and the user would run `module use ./modules`.

So with this PR, we are adding in the concept of views, and that is a view akin to a view in a database! So alongside this primarily modules directory, we might have:

```bash
 tree views/
views/
└── mpi
    ├── emacs
    │   └── 27.2.lua -> /home/vanessa/Desktop/Code/shpc/modules/ghcr.io/autamus/emacs/27.2/module.lua
    └── view.yaml
```
Where (as you see above) the module file is actually a symbolic link to the modules folder I showed above. This way, the admin can install smaller groups of module and just load one view, e.g.,:

```bash
module use views/mpi
```
or via shpc
```bash
module use $(shpc view mpi)
```
So the current question we are asking is - how do we further customize a view? Our current approach would customize the base module file. But here since we have the symlink, we can't do that without editing ALL of the other ones. So if there were no other option, we'd be forced to write another custom file there, which kind of defeats the purpose of the view and symlink.

So my question above is if there is any clever way, to say have a file one level up in a view, e.g.,:

```bash
 tree views/
views/
└── mpi
    ├── <customizations-always-loaded
    ├── emacs
    │   └── 27.2.lua -> /home/vanessa/Desktop/Code/shpc/modules/ghcr.io/autamus/emacs/27.2/module.lua
    └── view.yaml
```
that could be guaranteed to load (with some trick) always when the user does module use views/mpi?

> You may use the [ModulesCurrentModulefile](https://modules.readthedocs.io/en/latest/modulefile.html#mfvar-ModulesCurrentModulefile) variable within the modulefile to make it aware of its own fullpath name. Then adapt the evaluation, to define specific requirements based on this value.

This is an interesting idea, would it work to provide the correct path as the symlink? If so, if we could do something like:

- get custom path
- look up one level for a specific named thing
- if it exists, load it (the customizations) if not, no action (the regular install)

Or in more abstract terms, I think we want some ability to control children modules based on some file in a parent directory. E.g.,

```
A/
   b.file
   c/
      d.file
```
Where everything in c (d.file) is customized based on the presence of b.file. Sorry if that's not clear, it's after 2am here and I should not still be up, hehe. I can try again to explain more / answer more questions tomorrow... err later today :laughing:

marcodelapierre:
Here's one possibility, for Lmod only:
https://lmod.readthedocs.io/en/latest/050_lua_modulefiles.html

We could add a `try_load` statement in all modules. This is a special module load, which fails silently if broken.
For each view, we would then have a hidden module at the root of the tree, with the extra customisations:
```
$ ls * views/mpi

openfoam/1.0
openfoam/2.0
lammps/21
.view_module
```

and then in the modulefiles: 
```
try_load(".view_module")
```

Not sure what the equivalent in TCL modules is for `try_load`.

marcodelapierre:
....and it looks like there is!

https://lmod.readthedocs.io/en/latest/051_tcl_modulefiles.html

`module try-load`

vsoch:
Ding ding winner winner.., chicken dinner!! 🐔 

will test this out tomorrow! If y’all want to brainstorm what that file might look like for each while I’m snoozing will be greatly appreciated help ❤️ Night!

marcodelapierre:
Well, that extra module may open many doors....
The first that come to mind  (just copy-pasting from above):
* extra module loads (eg MPI related)
* setting values for `SINGULARITY_COMMAND_OPTS`, eg GPU flags

muffato:
`try-load` is a great idea ! So basically this `.view_module` file would contain
```
module load A
module load B
...
```
mirroring the list of system modules from the `view.yaml` ?

vsoch:
@muffato yes! and any further customizations we want for the view in the future, but right now let's scope to just loading extra modules.

shpc/main/modules/views.py
```diff
@@ -14,15 +15,50 @@
 import sys
 
 
+# Supported variables and defaults
+supported_view_variables = {"system_modules": []}
+
+
+class ViewModule:
+    """
+    A VersionModule is a .view_module written to the base of a view.
```
muffato, line 24:
```suggestion
    A ViewModule is a .view_module written to the base of a view.
```

vsoch:
okay looks like try-load was just implemented in early 2021: https://github.com/cea-hpc/modules/issues/392 so we are going to hit cases where it doesn't work. I think I will need to add a condition to check the version and only try that command if the release is later than that, brb.

muffato:
> okay looks like try-load was just implemented in early 2021: [cea-hpc/modules#392](https://github.com/cea-hpc/modules/issues/392) so we are going to hit cases where it doesn't work. I think I will need to add a condition to check the version and only try that command if the release is later than that, brb.

Good catch ! Indeed, on an older version of TCL modules (e.g. 4.3.0 2019-07-26) it fails
```
  Module ERROR: module: Invalid command 'try-load'
```

vsoch:
okay looks like I broke some logic for version files - looking at that now.

vsoch:
okay, huzzah everything working, now ready for review! Way to go me for introducing a change in the name of a variable and giving myself more errors to debug. :laughing:  So the changes just pushed include:

- support for `shpc view add mpi <var> <names>...` where (for now) the only supported thing is to add `system_modules` but you can imagine we expand this to other things. As an example (from the docs):

```bash
$ shpc view add mpi system_modules openmpi mymod
Wrote updated .view_module: /home/vanessa/Desktop/Code/shpc/views/mpi/.view_module
```
And I'm just realizing we probably need the same functionality but for "remove" - do y'all agree? I can add that after dinner if yes.

- basic tests for this, although I don't do any testing in the GitHub workflows to actually create some kind of system module and then test that it's loaded in the view (if anyone has suggestions here would be appreciated!) We do test in the test_views.py just that the file is generated, etc.
- adding the view_module for each of tcl/lua, with the exception that we can't use this approach for modules < version 4.8. I had to add a check for that in the tcl file, and also in the documentation as a warning to users. They can use views, but won't be able to customize with this approach.
- I expanded our GitHub testing matrix to take an additional module software version, so we can be sure that the try-load doesn't result in an error for both cases.

vsoch:
okay added support for remove (and better tests for both) hopefully I didn't break anything!

https://github.com/singularityhub/singularity-hpc/pull/545/commits/c17d20b3c0e7258a56a955a8b016107fbb08c95d

vsoch:
This is what I have now:

```tcl
# https://github.com/cea-hpc/modules/issues/392
if {[info exists ModuleTool] && $ModuleTool eq {Modules}
    && [versioncmp $ModuleToolVersion 4.8] >= 0} {
    module try-load .view_module
}
```
It won't fail, but it won't load the module (so it won't work for these older versions). I think I'm okay with that - views will still work (just without customization) and it will be less likely as we move forward and people update their old modules!

marcodelapierre:
> it's only under the view folder, so how would that happen? We'd just need a good way to unload the entire view folder (isn't that already a thing?)

@vsoch, this is the situation I have in mind.
As you say, views are practically enabled/disabled by issuing `module use`/`module unuse` to the corresponding ` view root directory. What these commands do is to add/remove the directory path from the colon sep list in the variable `MODULEPATH`. Typically, this variable will contain multiple paths.
What I was imagining is a situation where multiple views might be activated at the same time. This is where I see the problem arising.
In my deployments, I wouldn't have that, but...
I imagine Georgia's scenario, where she needs a view for MPICH, one for OpenMPI, and most likely...one for non-MPI code.
So I imagine users in her centre would always have the non-MPI view enabled, and then when loading eg the MPICH module, they would ALSO get the MPICH view activated. And that's when the issue manifests.

vsoch:
> What I was imagining is a situation where multiple views might be activated at the same time. This is where I see the problem arising.

I think (at least the way I envisioned views) we would discourage this - the idea of a view is that one is loaded at once. I wonder if we should even provide additional support to ensure the others are unloaded? 

E.g., the reason this is problematic (in addition to the above ) is that you can  have duplicate modules across views. If someone really needs to load two views they probably should have made just one view with what they need.

But yeah that's a good point - can you play around with some syntax and see if you can get your custom name idea working?

vsoch:
Yeah I think I agree - it's likely people won't do what we suggest anyway - I remember being a module user and I was very good at breaking things!!

> OK, I will see if I can flash out a syntax to protect from these things anyway :)

Woot!

marcodelapierre:
> Yeah I think I agree - it's likely people won't do what we suggest anyway - I remember being a module user and I was very good at breaking things!!

Ahah exactly! Same thing here...I was one of the the sys admins nightmares lol

vsoch:
> Ahah exactly! Same thing here...I was one of the the sys admins nightmares lol

lol! Did you also set up little web servers and play a game of cat / mouse for them to find them and take them down? That was my specialty :mouse: :cheese: :cat:

marcodelapierre:
> > Ahah exactly! Same thing here...I was one of the the sys admins nightmares lol
> 
> lol! Did you also set up little web servers and play a game of cat / mouse for them to find them and take them down? That was my specialty 🐭 🧀 🐱

Lol 😂 my specialty was ... parallel filesystem piracy 🏴‍☠️

xdelaruelle:
> Is there a standard way in Environment Modules, to disable error messages for failing `module load`s issued inside a modulefile?
> For context, what we're doing here is adding a blind `try-load <some_module>` in all modulefiles, but then `<some_module>` will exist only in certain cases.

You may use [is-avail](https://modules.readthedocs.io/en/latest/modulefile.html#mfcmd-is-avail) to test if modulefile exist prior loading it:

```tcl
if {[is-avail <some_module>]} {
    module load <some_module>
}
```

`is-avail` is there since version 4.1 (2018). It was added recently on Lmod (2021 I think).

But based on what you describe, I would suggest to `source` the `.view_module` file directly, and its code will be interpreted as if it were part of the module called by the user. `source` is a Tcl command, this is not specific to Modules. So you just need to check that `.view_module` exist as a file, not as a modulefile, which is simpler.

A quick example:

```console
$ cat foo
#%Module

set dir [file dirname $ModulesCurrentModulefile]
if {[file exists $dir/.view_modules]} {
    source $dir/.view_modules
}
$ mkdir mp
$ ln -s ../foo mp/foo
$ echo "module load bar" >mp/.view_modules
$ echo "#%Module" >mp/bar
$ module use ./mp
$ ml foo
Loading foo
  Loading requirement: bar
[xa@mosaic modules]$ ml
Currently Loaded Modulefiles:
 1) bar   2) foo
```

Also as a result, you do not see a `.view_modules` module in the loaded module list.

marcodelapierre:
Many thanks Xavier, great input!

However, if sourcing, when the module is unloaded its changes will not be undone.
To overcome this, we could mix the two approaches as in:
```
if {[file exists $dir/.view_modules]} {
    module load .view_modules
}
```
As you suggest, this syntax would work with all Env Modules versions.
The only little price is that the module WILL be in the `module list`, but I think it's fair enough, its changes are indeed in the shell environment.

vsoch:
@marcodelapierre would $dir get the right path (to the view) with the symbolic link?

xdelaruelle:
> However, if sourcing, when the module is unloaded its changes will not be undone.

`source` is also evaluated in `unload` mode. The content of sourced file is interpreted in unload mode, so if there are `module load` commands in this sourced file to load specific requirements for the view, they will become `module unload` when main module will be unloaded.

marcodelapierre:
> @marcodelapierre would $dir get the right path (to the view) with the symbolic link?

to be tested I reckon, it's in my list :)

marcodelapierre:
> > However, if sourcing, when the module is unloaded its changes will not be undone.
> 
> `source` is also evaluated in `unload` mode. The content of sourced file is interpreted in unload mode, so if there are `module load` commands in this sourced file to load specific requirements for the view, they will become `module unload` when main module will be unloaded.

Thanks @xdelaruelle ! Honestly, I was confusing this TCL `source` with a system Bash sourcing. Sounds great then!

muffato:
Thanks @xdelaruelle , great idea !
@vsoch and @marcodelapierre , re. users only having a single view active at a time. Another use-case I imagine for views is having per-team or per-project views, so a way of segmenting the module tree to declutter and structure the output of `module avail`.
Using the full path `$dir/.view_modules` should address this ?

shpc/main/modules/views.py
```diff
@@ -0,0 +1,504 @@
...
+            if force:
+                logger.info(
+                    "Overwriting view %s install of %s, as requested"
+                    % (self.name, module_dir)
+                )
+            elif not utils.confirm_action(
+                "%s in view %s already exists, are you sure you want to overwrite"
+                % (self._sub_module_base(module_dir), self.name)
+            ):
+                sys.exit(0)
```
muffato, line 342: Hi @vsoch . When re-installing a module that's already been installed, `shpc` asks the user to confirm whether they want to overwrite the symlink (code highlighted here), but not whether they want to replace the actual module itself. Don't you think it should first (and perhaps only) ask about the module ?

vsoch: The assumption is that re-installing is fairly harmless, it would regenerate "the same" file. A symlink, on the other hand, could be a totally different module with the same name.

vsoch:
> Using the full path $dir/.view_modules should address this ?

Possibly but we need to test if the symbolic link uses the actual or linked path.

marcodelapierre:
Hey folks, did some TCL/Lua tests, and can confirm that when looking for files, the symlink path by default is kept, and not traversed to get the original one.

Good!

So, with TCL we're essentially good to go, as this works, as suggested by @xdelaruelle , including with symlinks:

```
set view_dir "[file dirname [file dirname ${ModulesCurrentModulefile}] ]"
set view_module ".view_module"
set view_modulefile "${view_dir}/${view_module}"

if {[file exists ${view_modulefile}]} {
    source ${view_modulefile}
}
```

The story for Lua is a bit more complicated, next comment.

marcodelapierre:
Now to Lua/Lmod.

Symlinked paths are still handled correctly, however .. I could not find the Lua equivalent of TCL `source`, e.g. a command that allows the source/import another Lua script, given the full file path of the script.

Hence, what I thought was still to use traditional module load, while enforcing a view filename whose name is view specific, as to avoid the risk of inadvertently loading the wrong view file (when there are multiple views in MODULEPATH).

Again, note the underlying assumption that a view tree has the fixed general structure: $views_base/<tool>/<version-modulefile>.

So in Lua:
```
local view_dir = string.gsub(myFileName():match("(.*[/])") or ".","/%a*/$","")
local view_name = string.gsub(view_dir,".*/","")
local view_module = '.view_' .. view_name
local view_modulefile = view_dir .. '/' .. view_module .. '.lua'

if isFile(view_modulefile) then
  load(view_module)
end
```

As a result of these two implementations, there is a slight difference, as TCL does not "load" an extra module in the module list, whereas Lmod does.
We can make these two consistent if:
1. we find a Lua way of "sourcing" a Lua script from its full path; or 
2. we also use module load in TCL

marcodelapierre:
Hope this can help :-)

vsoch:
What about `source_sh`? https://lmod.readthedocs.io/en/latest/260_sh_to_modulefile.html

marcodelapierre:
`source_sh` is super-new in Lmod (version 8.6), so I am afraid it would not work in a lot of sites..

vsoch:
okay then I'm not sure what is best to do - let's see if others have opinions.

vsoch:
Hey folks! Wanted to check in to see if others had opinions about the best approach to take?

muffato:
Hi @vsoch and @marcodelapierre . (replying here rather than in #549). I haven't had time in the past month to follow up on any of the open issues / PRs.

I just tried https://github.com/singularityhub/singularity-hpc/pull/545#issuecomment-1114491377 (`source` on TCL) on my setup and I can confirm it works 💪🏼 . The "system modules" are added as necessary *and* removed as well during `module unload`. However, as opposite to what you claim @marcodelapierre in https://github.com/singularityhub/singularity-hpc/pull/545#issuecomment-1114501361 , they seem to be properly loaded. See below: `htstools/1.15.1` is a regular module from the view, `ISG/s3cmd/2.2.0` is a system module of the view

```
$ module list
No Modulefiles Currently Loaded.
$ echo $LOADEDMODULES

$ module load htstools/1.15.1
Loading htstools/1.15.1
  Loading requirement: ISG/s3cmd/2.2.0
$ module list
Currently Loaded Modulefiles:
 1) ISG/s3cmd/2.2.0   2) htstools/1.15.1  
$ echo $LOADEDMODULES
ISG/s3cmd/2.2.0:htstools/1.15.1
$ module unload htstools/1.15.1 
Unloading htstools/1.15.1
  Unloading useless requirement: ISG/s3cmd/2.2.0
```

(I'm on _Modules Release 4.3.0 (2019-07-26)_)

-----
Then, about the fact old Lmod is not able to source a file by its absolute path, I think it's reasonable to expect regular modules not to be named `.view_<view-name>`.

vsoch:
okay so should we move forward with @marcodelapierre suggestions for the files?

marcodelapierre:
Thanks for the additional tests @muffato :-) 

On TCL, it's interesting to see how behaviour changes (and happily improves!) over versions.

I did my tests with
```
VERSION=3.2.10.6
DATE=2017-01-09
```

With this old one, the functionality of `source`'ing a TCL file works, but when loading and unloading. I confirm the only hiccup is that the sourced file is NOT in `module list`, nor in `$LOADEDMODULES`.
Not an issue at all, but good to aware of these slightly different behaviours.

vsoch:
okay, so I'm not hearing any strong resistance to implementing these new ideas into the PR! I need to get my head back to where I was over a month ago, but I'll try to make these updates in an evening sometime this week. Thanks for testing out!

vsoch:
oops I totally did that wrong, lol.

vsoch:
@marcodelapierre @muffato is this ready for merge?

marcodelapierre:
> I totally forgot about the includes directory! I think this is what we want? [3631e6d](https://github.com/singularityhub/singularity-hpc/commit/3631e6ddbc3aaac78b3d6223a819497abbf051d4)

Yes @vsoch I would say we need this commit embedded in this PR!

marcodelapierre:
> @marcodelapierre @muffato is this ready for merge?

Other than adding the commit above, it looks ready to me.
I will not have the chance to test on a cluster for a bit, but please go ahead :-)

vsoch:
@marcodelapierre the commit is already added to the PR - did I do something wrong?

marcodelapierre:
sorry @vsoch ... I checked out the branch without fetching, got a very old commit .... blushing in shame now lol! 

all good to go!

vsoch:
Ok going to merge, and then I will be able
to properly review the other PRs! I’ll hold off on releasing for a few days in case y’all find any last minute bugs, and then we can also get the other PRs in the latest release.

This particular feature has been a ton of collaborative work and I’m so thankful for you both! 🎉 This particular project continues to be fun and fulfilling and I’m grateful for that.

marcodelapierre:
Thank you guys! :-)

This specific functionality is likely to address requests that came in recent months from other people, which I am also very happy about

muffato:
Wonderful news. Thank you so much !

PR #548, opened 2022-05-02
Various wrapper scripts improvement

muffato:
Hi @vsoch 

First of all, let me explain why I'm making this pull-request. I've got a software that necessitates a different wrapper script than the default `singularity.sh`, and I expect that other software will require the same change. Therefore, instead of defining those at the container level (hardcoding the tool / command-line parameters), I would like to save the alternative `singularity.sh` wrapper scripts in a central location. shpc currently only allows a single template per container technology, so there isn't really room for that.

The purpose of this PR is therefore to allow defining alternative template scripts that can be selected for different tools. I hijack the `singularity_scripts` directive for this. If an entry in `singularity_scripts:` corresponds to an alias listed in `aliases:`, then the script is used instead of the default `singularity.sh`. The template scripts are searched 1) in the `container.yaml` directory, 2) the user-defined global template directory, 3) shpc's own global template directory.

For instance, in `container.yaml` I have
```
aliases:
  longranger: /usr/local/bin/longranger
singularity_scripts:
  longranger: singularity_with_lsf.sh
```
and `singularity_with_lsf.sh` lives in the global template directory I've defined in `shpc/settings.yml` under `wrapper_scripts.templates`. Importantly it requires variables such as `alias.command`, etc, to be defined, unlike the usual `singularity_scripts`.

I've realised _after_ having written the code that the `aliases` section has a "long" form where aliases can be defined like this:
```
aliases:
- name: python
  command: /usr/local/bin/python
  docker_options: -it
```
and I think I would now rather add an option there such `singularity_template: singularity_with_lsf.sh` instead of hijacking `singularity_scripts`.

At the same time, I've done a few modifications to the way the search path for templates is defined, so that files are not searched in the current directory.
I also removed the constraint that `podman.sh` scripts can't be used (they were turned to `docker.sh` by the code). If the user wants to name their template script `podman.sh`, why not allow it ?

The PR also includes a couple of minor changes in container/{docker,podman}.py to improve how the classes are connected.

Best,
Matthieu


shpc/main/container/docker.py
```diff
@@ -33,13 +33,19 @@ def __init__(self):
             )
         super(DockerContainer, self).__init__()
 
+    def get_shell_path(self):
```
vsoch, line 36: Perhaps shell_path as a property?
```suggestion
    @property
    def shell_path(self):
```

shpc/main/wrappers/__init__.py
```diff
@@ -22,40 +22,54 @@ def generate(image, container, config, **kwargs):
...
-            wrapper = get_wrapper_script(template_name)(
-                settings=settings,
-                wrapper_template=template_name,
-                config=config,
-                container=container,
-                image=image,
-                **kwargs
-            )
-            generated += wrapper.generate_aliases()
+    constructor_kwargs = {
```
vsoch, line 25: I think it would still be nice to have a get_wrapper_script function - e.g., it's different from generate (which is going to write it) but just goes through the logic to return the thing we intend to parse.  This also will make it easier to write tests for different scenarios of getting the script (which I think we likely need!)

shpc/main/wrappers/__init__.py
```diff
@@ -22,40 +22,54 @@ def generate(image, container, config, **kwargs):
...
+        "config": config,
+        "container": container,
+        "image": image,
+        **kwargs,
+    }
+
+    # Default wrapper for this container technology. Will be used for all
+    # aliases (unless overriden)
+    default_wrapper = None
+    default_template_name = settings.wrapper_scripts.get(container.command)
```
vsoch, line 36: I love how you refactored this, so much nicer looking! <3

shpc/main/wrappers/__init__.py
```diff
@@ -22,40 +22,54 @@ def generate(image, container, config, **kwargs):
...
-                config=config,
-                container=container,
-                image=image,
-                **kwargs
-            )
-            generated += wrapper.generate(alias)
+    # Additional commands defined in the custom container.yaml script section
+    container_templates = scripts.get(container.templatefile, [])
+    container_wrappers = {}
+    for alias, template in container_templates.items():
```
vsoch, line 53: Ah I think I see your confusion here - we have the idea of an "alias" as in "the name of the command we are writing a wrapper for instead (e.g., fork)" and then the aliases proper, which the module templates use that read in the aliases (and map to a common structure) to render into the template.

vsoch: It might be helpful to update the docs in places where I didn't do a good job of explaining what to expect.

vsoch:
@muffato bravo! This is seriously beautiful code, it's a huge improvement over the current design! Let me respond with a few thoughts and suggestions.

> The purpose of this PR is therefore to allow defining alternative template scripts that can be selected for different tools. I hijack the singularity_scripts directive for this. If an entry in singularity_scripts: corresponds to an alias listed in aliases:, then the script is used instead of the default singularity.sh. 

Could there be a case where there is a common alias and the template script is defined and it is accidentally used? It sounds like you are relying on the container.yaml for the customization, and I'm wondering if we need to put our heads together and think about how this could better be controlled. E.g., is this something that would make sense to define for a view? Should there be an install flag that is a one off "use this wrapper script for this alias?" or some kind of customization file we allow to apply? I think the changes here are good/needed and let's keep thinking about this case of wanting not just customizations for a group of containers, but for a specific container. 

> The template scripts are searched 1) in the container.yaml directory, 2) the user-defined global template directory, 3) shpc's own global template directory.

Yes I like this design much better.

> and singularity_with_lsf.sh lives in the global template directory I've defined in shpc/settings.yml under wrapper_scripts.templates. Importantly it requires variables such as alias.command, etc, to be defined, unlike the usual singularity_scripts.

Will there be an error if something is not? Or more generally, do we have a way to enforce what is needed and alert the user if something is missing?

> I've realised after having written the code that the aliases section has a "long" form where aliases can be defined like this: and I think I would now rather add an option there such singularity_template: singularity_with_lsf.sh instead of hijacking singularity_scripts.

Yes indeed - this custom format is mapped into the same alias structure (but has more metadata). For your use case, is the wrapper something you think should be added to the registry (and thus added to this alias options section) or it's a one off for your specific cluster? If the latter (to go back to my comment above) I wonder how this could be an easy workflow. Right now you are saying:

1. Write a custom wrapper
2. Update the container.yaml with it (likely not something you can contribute to the registry)
3. Install

Or if it's the case that it's a general script that could be contributed, should we have the wrapper aliases be a different alias? E.g., for your example:
```yaml
aliases:
  longranger: /usr/local/bin/longranger
singularity_scripts:
  longranger-lsf: singularity_with_lsf.sh
```
Or something of that nature? I could imagine something like:

```yaml
aliases:
- name: longranger-lsf
  singularity_script: singularity_with_lsf.sh
  description: This is the longranger command as a wrapper intended for running with lsf
```
And that way we could package a bunch of wrapper scripts with a container (for different scenarios) that would be installed, but sort of optional to use. Or maybe we could allow the person to skip aliases, e.g.:

```bash
$ shpc install <module> --skip-alias longranger-lsf
```
If they have strong feelings about not including something :) But it's a fairly small addition so (personally speaking) I wouldn't have issue. And I haven't thought this out fully yet, but if there is some case where you want to apply a wrapper script more globally to all modules in a view, we could have some way to do that. Or perhaps in the case you don't want to edit the container.yaml but you do want a wrapper:

```bash
$ shpc install <module> --wrap longranger singularity_with_lsf.sh
```

The above would:
1. determine your container technology to derive if you want singularity_script, docker_script, or podman_script.
2. dynamically add the alias for longranger and to be associated with the wrapper
3. continue the install as if it had been already in the container.yaml to generate the wrapper!

The issue there would be reproducibility - e.g., if your install tree blows up I hope you saved all the commands :)

So overall I think let's focus on the following:

- it would be really good to be able to test this better, specifically what script is retrieved with different settings. There are examples of tests that "one off" settings, and perhaps if we have a get_wrapper_script function (I know you removed it but I think it would be useful for this testing scenario) those tests would be easy to write!
- Off of that point, I do think we should have proper tests here, at least more than I added because I totally missed some of the details you are fixing here!
- Perhaps we can extend the alias definition (the longer form) above to have a "singularity_script" or "docker_script" and then a description for it?
- And if you agree with the above, the "would be nice" would be the command line updates - to allow disabling a one off wrapper script install, OR enable one, one off.

muffato:
TODO list:
- [ ] it would be really good to be able to test this better, specifically what script is retrieved with different settings. There are examples of tests that "one off" settings, and perhaps if we have a get_wrapper_script function (I know you removed it but I think it would be useful for this testing scenario) those tests would be easy to write!
- [ ] Off of that point, I do think we should have proper tests here, at least more than I added because I totally missed some of the details you are fixing here!
- [x] Perhaps we can extend the alias definition (the longer form) above to have a "singularity_script" or "docker_script" and then a description for it?
- [ ] And if you agree with the above, the "would be nice" would be the command line updates - to allow disabling a one off wrapper script install, OR enable one, one off.
- [ ] And for all of the above, documentation updates to make things more clear! I think your struggle with figuring out what I originally wrote is indication that it could be improved!

shpc/main/wrappers/__init__.py
muffato: I've given this a shot in the latest commits (cf `find_wrapper_script` in `base.py`). The function should be easy to test as it doesn't require any of the Jinja stuff of `load_template`. Let me know if that's what you had in mind. Then I can write the tests / doc

vsoch:
I will be able to review again tomorrow!

shpc/main/wrappers/__init__.py
muffato: > 1. Why do we set `self.wrapper_template` and then provide it to the function? It should probably be one or the other - either set it (and use it in the function without providing as an argument) or provide to the function and allow it to be set there.

Ah yes, I've gone back and forth about this and I can see there are remnants of both approaches. Initially I wanted to  make it an argument of the function because it's not much used in other places, but then I thought it'd be good for the class to keep track of the most important stuff, hence the `self.`.

_But_ I've also been thinking about the purpose of this module `shpc/main/wrappers/base.py`. It has a single class, `WrapperScript`, and by making `wrapper_template` a function argument, the class essentially becomes a holder of settings, configuration, overall parameters, etc, – nothing alias-specific –, that has methods to find and load templates, and generate the scripts. And all the alias-specific stuff would be handled as _parameters_ of these methods only. Therefore, there wouldn't be a need to instantiate a new `WrapperScript` for each alias / wrapper script in `shpc/main/wrappers/__init__.py`. I may then rename the class to say `WrapperScriptGenerator`, and the file to say `shpc/main/wrappers/generator.py`, and instantiate it just once in `shpc/main/wrappers/__init__.py:generate`.

What would you think of that ?

Otherwise, I may come back to the opposite approach, and reinstate `wrapper_template` as a mandatory constructor argument 

> 2. Does `self.template_file` need to be used anywhere, other than providing the directory path at the end of the function?

It's its sole use, and I've already made it a local variable instead in a commit I haven't pushed yet.

> 3. if we are returning the template directories (and not the wrapper script, which is set) should it be called `set_wrapper_script`?

I'll probably change the name indeed, also depending of 1) and 4)

> 4. I usually like one function to do one thing - is there a way to piece apart finding the files vs. the directories?

Absolutely

muffato: Mostly done (I still need to test that it works 😅)

muffato:
Hi @vsoch . I need a couple of pointers to add the unit tests:
- Does the pytest stuff run on GitHub CI ? I don't see where / how
- In order to test this, I'd need test wrapper templates, test container.yaml, and etc. How would I go about adding those, but also changing the settings, like paths etc, for the tests ?

Matthieu

vsoch:
Looks like pytest was running on circle, and stopped at some point (I don’t see it running recently). If you want to take a shot at adding to a GitHub recipe go for it (maybe main.yaml) otherwise I can get this added back asap. https://github.com/singularityhub/singularity-hpc/blob/eaf978cfe2b4e8ac2a3b9d141cff9ba79aaae69b/.circleci/config.yml#L98

And then for test data just add to an appropriate test data folder. And see this example https://github.com/singularityhub/singularity-hpc/blob/eaf978cfe2b4e8ac2a3b9d141cff9ba79aaae69b/shpc/tests/helpers.py#L36 for creating a test client and changing a config value on the fly.

I’ll take a look into circle soon!

vsoch:
Ok found the issue! And I’m not a nutter because circle has been running, just not for forks. I just clicked this so hopefully new pushes will run! Thanks for catching this! 
A55C29E6-861F-4445-BB33-FCBB14F3C1A0

shpc/tests/test_wrappers.py
```diff
@@ -0,0 +1,118 @@
...
+    # Note: the module_sys and container_tech parameters are actually not used in the test
+    client = init_client(str(tmp_path), "tcl", "singularity")
+    # The file name is actually not used in the test, only its enclosing directory
+    testdata_dir = os.path.join(here, "testdata")
+    container_yaml = os.path.join(testdata_dir, "quay-container.yaml")
+    config = container.ContainerConfig(container_yaml)
+    ws = wrappers_base.WrapperScript(
+        "dummy.sh",  # Not used in this test
+        client.settings,
+        None,  # image
```
vsoch, line 42: nit: do these not have defaults of None? If we use kwargs for the config do we need this here?

shpc/tests/test_wrappers.py
```diff
@@ -0,0 +1,118 @@
...
+    [
+        (False, False),
+        (False, True),
+        (True, False),
+        (True, True),
+    ],
+)
+def test_find_wrapper_script(tmp_path, make_absolute, exists):
+    """
+    Test views:
```
vsoch, line 81: Should this comment be test (something else?)

vsoch:
woohoo the circle tests ran!

vsoch:
@muffato I have some nits above, but it's up to you if you want to change them - I'm happy with the current PR and ready to merge! Let me know your preference.

shpc/tests/test_wrappers.py
muffato: Yep. Forgot to change it

shpc/tests/test_wrappers.py
muffato: I fixed the `None` of the "container" parameter but I can't directly get rid of the `None` of the "image" parameter because it's a positional argument, not a keyword one. As a solution, I could pass the first two as keywords, or turn "image" to a keyword one

vsoch: Don’t worry about that - your change here should be good!

vsoch: If we were to expose this class externally we could maybe think harder about this, but since it’s an internal one I don’t think we need to go further than what we have for the test.

muffato:
Then we're good to go !

vsoch (approved):
😎🥳🎉

PR #549, opened 2022-05-27
singularity modulefile templates: add SINGULARITY_CONTAINER shell variable

marcodelapierre:
Hi @vsoch , as per title:

adding the variable `SINGULARITY_SIF` for singularity modulefile templates, for user cases where people need to use the sif file.
Our practical use case is: migrating users from old workflows using SIF paths, onto modern (!) SHPC-based workflows.


vsoch:
hey @marcodelapierre ! I can definitely review this soon - do you have an opinion about what we should do with #545 ? It seems like we had some momentum and then needed to make a final decision and there was radio silence. I want to be careful about adding things on top of that because it might be much harder to resolve merge conflicts. Is the conclusion that adding views isn't a good idea / wanted enough at this time?

marcodelapierre:
On 545, I had the impression that the feature was considered powerful and useful from multiple parties. It would be good to finalise.
I can't recall if there were outstanding points from other people.
Regarding the one I raised on loading the hidden modules, I think we have a working implementation for both Tcl and Lua (my latest comments there), which uses "old enough" syntax to be reasonably back-compatible. It would need implementation and some minimal testing. I think it was good because it allowed to enforce some of the desired functionalities for views, without major disadvantages. But if you/others have any concerns, let's definitely take them into account.
More broadly, what is your perception on the status of the PR?
Thanks, have a good.....evening? :)

vsoch:
> More broadly, what is your perception on the status of the PR?

I think we are _so_ close, there has been a lot of thinking and discussion, and it's really that final bit that (others, primarily not you) might have commented on. My bias: I think I would be OK with implementing something "forward thinking" in the sense that super old versions of some module software just wouldn't be able to use it. It's sort of a backhanded way to encourage them to update? :imp: I think I was hoping for blessing from others in the thread, but given silence, we should go ahead with your suggestion! I'll put that on my TODO for after I finish what I'm currently working on, either later tonight or tomorrow.

> Thanks, have a good.....evening? :)

Haha indeed! It's a little before 9pm here. But the night is young! _howls at the moon and scurries into the darkness_

marcodelapierre:
OK, thanks!
On the "old" Lmod, I also like the approach of encouraging adoption of contemporary tools and versions. 

Unfortunately in this case I think it is risky, based on my (very specific) experience with Cray Supercomputers. In this context, the version of the module system is part of the Cray Software Environment, on which the HPC centre staff has limited action. 
Eg this is the Lmod version on our next Cray supercomputer, which we're configuring as we speak:
```
$ module --version

Modules based on Lua: Version 8.3.1  2020-02-16 19:46 :z
```

muffato:
Some mixed comments about this PR:
 - Have you considered setting `SINGULARITY_CONTAINER` rather than `SINGULARITY_SIF` to match the environment variable that singularity itself sets within the container ?
 - If someone loads another shpc module, then `SINGULARITY_SIF` will be overridden. Have you considered a function with a unique name, e.g.
    ```
    set-alias {|module_name|}-sif "echo ${containerPath}"
    ```

Related question as I check the diff of this PR: what is the use of the `SINGULARITY_SHELL` environment variable ? I can't see where it is used, as all shell references seem to be to `{{settings.singularity_shell}}`

vsoch:
@muffato I think `SINGULARITY_SHELL` is respected by Singularity itself, e.g., see https://sylabs.io/guides/3.7/user-guide/appendix.html?highlight=singularity_shell

marcodelapierre:
Thanks @muffato , I agree with both of your suggestions, going to implement them shortly, and commit to this PR

marcodelapierre:
Update following @muffato 's feedback:
* `SINGULARITY_SIF` variable renamed to `SINGULARITY_CONTAINER`, in compliance with Singularity conventions
* added shell function `|module_name|-container`, to provide an interface to the container name, that is unique to each container module

muffato:
@vsoch . I think this use case is already covered by `-s {{ settings.singularity_shell }}` in the `singularity shell` command at https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/modules/templates/singularity.tcl#L75 ?

@marcodelapierre : 🤩 . I didn't mean to force you to implement those asap ! so great !!

vsoch:
Thanks to you both! I’ll review this and #548 as soon as we get the add/views PR merged! That one is most complex (and longest worked on) and I think should go in first.

vsoch:
@marcodelapierre I'll be able to review again tomorrow!

vsoch (commented):
This looks good! Could we possibly add this environment variable to be documented in the user or developer docs somewhere?

marcodelapierre:
Just added mentions in the documentation, have a look :)

.vscode/tasks.json
```diff
@@ -0,0 +1,70 @@
+{
```
vsoch, line 1: oops! I don't think you meant to add this vscode file, although it's kind of fun to look at (I don't use vscode but I can see how it works a bit now!)

docs/getting_started/use-cases.rst
```diff
@@ -167,7 +167,7 @@ loaded, just do:
        This module is a singularity container wrapper for python v3.9.2-slim
        Commands include:
         - python-shell:
-           singularity shell -s /bin/bash /usr/WS2/sochat1/singularity-hpc/modules/python/3.9.2-slim/python-3.9.2-slim-    sha256:85ed629e6ff79d0bf796339ea188c863048e9aedbf7f946171266671ee5c04ef.sif
+           singularity shell -s /bin/bash /usr/WS2/sochat1/singularity-hpc/modules/python/3.9.2-slim/python-3.9.2-slim-sha256:85ed629e6ff79d0bf796339ea188c863048e9aedbf7f946171266671ee5c04ef.sif
```
vsoch, line 170: And I've made these changes to main so it's a generic path!
```suggestion
           singularity shell -s /bin/bash /home/shpc-user/singularity-hpc/modules/python/3.9.2-slim/python-3.9.2-slim-sha256:85ed629e6ff79d0bf796339ea188c863048e9aedbf7f946171266671ee5c04ef.sif
```

docs/getting_started/use-cases.rst
```diff
@@ -192,6 +192,17 @@ singularity recipe) then you can do:
     python/3.9.2-slim/module             : org.label-schema.usage.singularity.version  : 3.7.1-1.el7
 
 
+If your workflow requires knowledge of the local path to the sif image, this information
+can be output by using the "container" suffixed alias:
+
+.. code-block:: console
+
+    $ python-container
+    /usr/WS2/sochat1/singularity-hpc/modules/python/3.9.2-slim/python-3.9.2-slim-sha256:85ed629e6ff79d0bf796339ea188c863048e9aedbf7f946171266671ee5c04ef.sif
```
vsoch, line 201:
```suggestion
    /home/shpc-user/singularity-hpc/modules/python/3.9.2-slim/python-3.9.2-slim-sha256:85ed629e6ff79d0bf796339ea188c863048e9aedbf7f946171266671ee5c04ef.sif
```

vsoch (commented):
This is looking good! I think one more question - should we add a test to run this alias and then assert the output exists / is a sif extension (or something along those lines?)

.vscode/tasks.json
marcodelapierre: ahah funny indeed, didn't notice.
I didn't write it, it's the application that creates it for its own wizardries

marcodelapierre:
I think your proposed tests are good, we could test for the `|tool|-container` alias; the variable has the same contents, I don't see it going wrong if the alias succeeds.
As you propose:
* is there some output? 
* does it start with `/` and finish with `.sif`?

Not sure when I have time to look at this,  by any chance could you have a look? (if no, I will see....hopefully next week)

marcodelapierre:
otherwise I have implemented your other suggestions! :)

vsoch:
> Not sure when I have time to look at this, by any chance could you have a look? (if no, I will see....hopefully next week)

yes absolutely! I'll see if I can write the tests this evening.

vsoch:
I'll take a shot at finishing this up at the end of the work day!

PR #553, opened 2022-06-04
Add/sif variable for singularity

vsoch:
This will close #549 if it works here - I think this was the last detail todo!


vsoch:
@marcodelapierre I'm not sure what I'm doing wrong, but I can't get it work. :/ So I think I'll need your help with this after all! Sorry I couldn't help.

marcodelapierre:
No worries @vsoch, I will have a look as soon as I can, hopefully in the next few days.

marcodelapierre:
Ok, got one!

The Lmod error (empty container alias for Tcsh )is because I forgot to add the second definition, which indeed is required in the case of Csh/Tcsh.

Have committed the fix on my branch: 
https://github.com/singularityhub/singularity-hpc/compare/add/sif_variable_for_singularity...marcodelapierre:add/sif_variable_for_singularity

vsoch:
okay - just added your commits here!

> shall we add a dedicated Docker/Podman alias + function, or ditch the corresponding tests.

Not until someone asks for it, imho.

marcodelapierre:
oh wow! I suspect .. you broke Github @vsoch!! 😆

interestingly, I can see the error message, e.g. tcl 4.7.1 docker ;  I am using Safari on a Mac. Or maybe they solved it meanwhile...?

vsoch:
oh no, that means it's me! I've tried Firefox and Brave, so I could try Chrome as my last resort. I think it might be related to something with my ip-address because my linkedin site has been down (again spinning) for about a week, and that's true for any computer or browser in my apartment. When I connect to work VPN (on a work computer) it magically works. Of course it's a catch 22 because I'm not allowed to work on personal projects on a work computer, so I couldn't.

This might be really annoying, but could you show me one of the errors so I can try fixing it?

marcodelapierre:
I think the test fails because it Singularity specific, i.e. it looks for a .sif extension:
```
grep --quiet '\.sif$' container_output
```
Not sure what's the best way to test for the Docker image specification.. the prefix can be `docker.io/`, `quay.io/`, `nvcr.io/`, and so on ... so maybe grepping for `\.io/` . 
However, this might fail as I gather it is just a convention, and private registries might have a different naming?
So perhaps for docker/podman, we can only test that the output of the `-container` command is not empty...?

vsoch:
Ah of course! Thank you @marcodelapierre ! Yes I can think about how to fix this.

Update: chrome did not work, nor did disabling ad blockers (which wouldn't make sense since I've had them for years but worth a shot).

vsoch:
okay I found an old mac (and turned it on) and I can see the logs (thank goodness!)

vsoch:
oh man I am epically failing! @marcodelapierre do you see anything that I'm missing?

vsoch:
oh hmm docker is working but not podman...

vsoch:
I think it's maybe as you suggested, that we cannnot get the URI perfectly given a hash or registry is added.

vsoch:
woohoo! Well, that was an adventure! :laughing:

marcodelapierre:
Thanks Vanessa!

vsoch:
Also @marcodelapierre I've recently bit the bullet and got my own cloud accounts (for development) so if you ever get any 'wouldn't it be cool if.." ideas working with GCP or AWS... you know who to ping!!

marcodelapierre:
Thanks for mentioning Vanessa! Mostly on-premises right now, but will keep it in mind :-)

PR #554, opened 2022-06-11
Extra commands

muffato:
Hi

When doing some tests, I noticed that these two containers have more useful commands than listed.

Cheers,
Matthieu


vsoch (approved):
Thanks for adding these! To derive the original list I (usually) did it programmatically so I suspect these are new executables in the container. We might run into issues where an older container is missing commands, but in this case it’s easier to deprecate those versions, or suggest the person uses an earlier recipe file.

Thank you @muffato !

PR #557, opened 2022-06-28
adding support for alias files

### LINKED ISSUE #536: Tricky containers: many executables across distinct versions
Hi @vsoch,

so after thinking that MPI + containers was a bumpy road, I met ... OpenFoam! (which is also an MPI package, though irrelevant here)

At Pawsey we build custom OpenFoam containers for our researchers, as the stock images from the developers only come with OpenMPI, and we need ...MPICH (eheh).
I have just updated the images for the upcoming supercomputer, and I was starting to write the SHPC recipes, when.... 

Two interesting things!
1. Each OpenFoam version comes with 300+ executables; well, I said, it will be a longer SHPC recipe....but...
2. Distinct OpenFoam versions ship with different sets of these 300+ executables (as in, some executable names are different)

So, how to tackle this? -- brainstorming welcome!

Let me share my early thoughts, staring from point 2.
So I was thinking, how about enabling a container recipe to have sections that can be specific to specific tags, e.g.:
```
docker: vanessa/salad
[..]
latest:
  2: [..]
tags:
  1: [..]
  2: [..]

1:
  aliases:
    salad: /code/salad1

2:
  aliases:
    salad: /code/salad2
```
Doesn't have to be only for aliases, I guess it could also be other properties such as features, envs...
As a side note, we'd need to make sure that the latest tag is handled properly (but I suspect it would work out of the box right?).

And then there's still point 1. left.
The specific case of OpenFoam would still imply a 3000+ lines long recipe, so I was wondering whether it would make sense to implement a second feature, to allow the recipe include other yamls in the same dir, or subdirs. In this way, the recipe contents can be chunked down into many files, for improved readability and maintenance.

What do you think?

vsoch:
An alias file is a yaml file "lookup" alongside a container.yaml that can be used to provide tag specific commands for a particular
container tag. I am also adding the two pawsey containers as an example, and (when the main CI passes) we can trigger the container test CI, as I've also updated those tests to assert that all provided alias files exist!

This will close #536 


vsoch:
Docs preview for alias files:

image

marcodelapierre:
I love how this looks like! 
Will give it a go as soon as I can

vsoch:
Also @marcodelapierre when you have some time it would be fun to brainstorm an idea that I’ve been toying with - I really like this idea of providing container transparency (wrt) aliases - it was my initial vision for the scientific filesystem (building the aliases into the entry point) and the interactions here. What I’ve been chewing on is how to move these aliases outside of shpc, meaning making them possible for any container technology to find and use. It could mean something like:

- a custom container build that allows for customization or automation detection and writing of aliases
- Saving into some simple metadata file
- Using Oras OCI registry as storage to upload as a related artifact
- A custom tool that can find the container and alias file, and “enable” it (and I’m not sure what that means yet)


At the most complex level you could imagine a registry just for aliases, and a tool like shpc would ping it to look up containers. And the simplest level is a command line tool that tries to bring them together in some context.

marcodelapierre:
I will definitely follow-up on this, I would love to brainstorm on it!
Can you make a dedicated github issue?

vsoch:
Yes!

muffato:
Hello !
This sounds like a great feature. I don't have the use case right now, but I'm sure it's going to come very soon. So very pleased to see this pull request.

One comment: could this be called say `overrides` rather than `alias_files`, and the content merged with `container.yaml`'s. This would allow replacing not only the aliases, but also the environment variables, which can equally change over time !

Best,
Matthieu

vsoch:
@muffato that's a great idea! Would it be okay to still be tag specific, and are there any fields we don't want to override?

muffato:
1. I think that's fine for now. It could be useful at some point to support a filter language, so that multiple tags could be matched to the same override file on the same line. Maybe, to avoid the confusion, it could written down as a "long form" and what you have now would be the "short form" (like how aliases are defined). I'd be happy keeping this for later, when the  use case actually materialises, though I've noticed you're more eager than me to be future-proof ;)
2. Looking at the `container.yaml` files, I guess we should allow overriding `docker_scripts:` and `singularity_scripts:` too. And I think that should be it

vsoch:
Ok sounds good! I should be able to update the PR this evening.

marcodelapierre:
Cool!
Will test with my OpenFoam user case (tag-specific aliases only) after the updates to PR

vsoch:
okay - updates to use "overrides" and for a specific set of fields should be in!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/main) (0.0.x)
+ - adding support for container.yaml alias_files (0.0.56)
```
muffato, line 17:
```suggestion
 - adding support for container.yaml overrides (0.0.56)
```

shpc/main/container/config.py
```diff
@@ -259,26 +304,29 @@ def get_envars(self):
         """
         return dict(self.env) if self.env else {}
 
-    def get_aliases(self):
+    def get_aliases(self, tag=None):
```
muffato, line 307: This might be a very dumb question: why is this function changed ? I can't see the `tag` parameter being used, and I don't understand the renames `aliases` → `loaded` and `self.aliases` → `aliases`.
Is it a leftover from the previous version where only aliases could be overridden ?

muffato (commented):
It looks amazingly lean and clean ! Nice one :) Just a couple of comments:

shpc/main/container/config.py
vsoch: This tag=None should be removed - that was an oversight! When the changes were scoped to aliases, we had lines here that would essentially load the file (and aliases) on demand. Now as overrids, that happens once at the beginning of the install (so we don't need the tag here). It was my oversight to not remove that (now unneeded) tag parameter.

vsoch:
I think tests were ok, and the containers just took too long to pull.

shpc/main/client.py
```diff
@@ -185,7 +190,7 @@ def cleanup(tmpdir):
             if not test_exec:
                 continue
 
-            for alias in config.get_aliases():
+            for alias in config.get_aliases(tag):
```
muffato, line 193:
```suggestion
            for alias in config.get_aliases():
```

muffato:
Hi @vsoch . Can I double-check something with you ? I can't manage to install a container any more, even on the main branch. It's not able to separate the tag from the container name
```
$ git show --oneline 
b1de8eed7 (HEAD -> main, origin/main, origin/HEAD) Merge pull request #556 from marcodelapierre/update/registry/quay.io_pawsey_python
$ git diff
diff --git a/shpc/settings.yml b/shpc/settings.yml
index 557695697..7dfdb106a 100644
--- a/shpc/settings.yml
+++ b/shpc/settings.yml
@@ -5,7 +5,7 @@
 # set a default module system
 # Currently lmod and tcl are supported. To request an additional system, 
-module_sys: lmod
+module_sys: tcl
 
 # config editor
 config_editor: vim
@@ -70,7 +70,7 @@ wrapper_shell: /bin/bash
 # If you want a custom wrapper script for a one-off container, define it there with the script
 # stored alongside the container.yaml.
 wrapper_scripts:
-  enabled: false
+  enabled: true
 
   # use for docker aliases (set to null to disable)
   docker: docker.sh
$ shpc install quay.io/biocontainers/samtools:1.14--hb421002_0
  File "/nfs/users/nfs_m/mm49/.conda/envs/shpc/bin/shpc", line 33, in <module>
    sys.exit(load_entry_point('singularity-hpc', 'console_scripts', 'shpc')())
  File "/nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/shpc/client/__init__.py", line 453, in run_shpc
  File "/nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/shpc/client/install.py", line 30, in main
    cli.install(
  File "/nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/shpc/main/modules/base.py", line 417, in install
    self.container.install(
  File "/nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/shpc/main/container/singularity.py", line 223, in install
    wrapper_scripts = shpc.main.wrappers.generate(
  File "/nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/shpc/main/wrappers/__init__.py", line 71, in generate
    for alias, template_name in listing.items():
AttributeError: 'NoneType' object has no attribute 'items'
```
Does it work for you ?

vsoch:
Reproduced - only when I turned on wrapper scripts! The bug was that it needed to default to an {} given that the value was present and None, so I changed to .get or {}.

muffato:
Thanks, it can now get installed, but overrides don't work. I'm installing the container like this
```
shpc install quay.io/biocontainers/samtools:1.15--h3843a85_0
```

I think it's because`tag` in `shpc/main/modules/base.py:install` is still None when `config.load_override_file(tag)` is called, and it's way below that function that the version is split from the name if there is a colon in it. Because of that `load_override_file` doesn't think that overrides need to be loaded. Does that make sense ?

vsoch:
yep I totally see that! Did you add an overrides file to that container you are testing?

muffato:
Yes, thanks for checking ;) the overrides files are there, and detected. Here is the content of `self.overrides` at the beginning of `shpc/main/container/config.py:load_override_file`:
```
ordereddict([('1.14--hb421002_0', '1.14--hb421002_0.yaml'), ('1.15--h3843a85_0', '1.15--h3843a85_0.yaml')])
```

vsoch:
@muffato can you give me the complete file set to reproduce? I can also make some test cases from them!

muffato:
Updated `container.yaml` and two overrides files: [samtools.zip](https://github.com/singularityhub/singularity-hpc/files/9023729/samtools.zip)

Unpack in `registry/quay.io/biocontainers/samtools/` and try installing the tags listed in the table below. Each will have a different number of aliases, or an environment variable:

| Tag | Number of aliases | `99-shpc.sh` |
| --- | --- | --- |
| 1.13--h8c37831_0 | 14 | no var |
| 1.14--hb421002_0 | 27 | no var |
| 1.15--h3843a85_0 | 14 | `REFPATH` |

muffato:
@vsoch : these overrides are not meant to make it to the registry. I wrote them just to test different scenarios

muffato:
Also, seeing that you introduced `version` earlier in  `shpc/main/modules/base.py:install`, should it not be used instead of `tag` in these lines too ?
```
        # Add a .version file to indicate the level of versioning
        self.versionfile.write(os.path.join(self.settings.module_base, uri), tag.name)
        # For Singularity this is a path, podman is a uri. If None is returned
        # there was an error and we cleanup
        if not container_path:
            container_path = self.container.registry_pull(
                module_dir, container_dir, config, tag
            )
```

vsoch:
Agree - and sorry for adding the overrides files to the actual container - I knew the envar wasn't real but I though the aliases might be! Will remove them after dinner (soon).

vsoch:
We can't use it here:

```
if not container_path:
            container_path = self.container.registry_pull(
                module_dir, container_dir, config, tag
            )
```
because tag needs to be the full tag with digest, etc. But we can use it one level up! This function is admittedly a bit messy - I am going to see if I can neaten it up, at least with respect to not having tag.name _and_ version!

vsoch:
okay I tried cleaning it up, and removed the extra overrides! Hopefully it looks better now.

muffato:
All good for me, I can confirm it all works as intended. Thanks !
Just one thing: you actually need to also revert `registry/quay.io/biocontainers/samtools/container.yaml` to what it was before (full list of aliases). I edited it that way to sort of make sure I wasn't picking up leftovers from previous installations.

marcodelapierre:
I will aim to do some quick testing on my end, too, next week.
However feel free to go ahead with the merge any way, as you prefer.

marcodelapierre:
my testing will be on the lines of singularity + lmod + wrapper scripts + overrides

muffato (approved):

marcodelapierre:
done with my testing, all good on my side!

PR #560, opened 2022-07-07
Add tensorflow to right location

surak:
All containers of the default registry can be found by a simple "shpc install <name>". For example, `shpc install python`. 

All of those work, except for Tensorflow - for which you have to do a `shpc install tensorflow/tensorflow`. I find it inconsistent with the other containers.

This merge request simply move tensorflow one level up, so it behaves the same as the other containers of this registry.


vsoch:
@surak the reason we can't do that is because the module namespace respects the container namespace. So shpc install tensorflow/tensorflow installs https://hub.docker.com/r/tensorflow/tensorflow/. If you want the shorter name to load, if you add tensorflow to a view it should just be `module load tensorflow`.

vsoch:
@surak I love that you came here to contribute this is NOT wasting my time! I hope you come back soon :)

PR #567, opened 2022-07-19
Adding support for sync

### LINKED ISSUE #566: Update of local registry with new recipes
The shpc update seems to update only existing recipes, but to have all the latest ones from you, one needs to get a newer version of shpc, correct? 

What I mean is newer recipes which would not show on `shpc show` until you get a newer shpc.

Would it make sense to have a "update" setting which would bring new recipes as well? What do you think?

vsoch:
This still needs tests written - I think I will need to make an external remote registry with some arbitrary change to test, OR just test the interaction with the remote provider and handle the updates via local test files.

This also creates the concept of a registry, e.g., filesystem or remote that we can expand upon.

This will close #566 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/main) (0.0.x)
+ - Add support for upgrade (add new recipes) and update --all (0.0.58)
  - fixing bug to config set/get nested fields, default for wrapper scripts true (0.0.57)
   - change in behavior, instead of `key:value` for set/add `key value` now also works.
  - adding support for container.yaml overrides (0.0.56)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/main) (0.0.x)
- - Add support for upgrade (add new recipes) and update --all (0.0.58)
+ - Add support for sync-registry (add new recipes) and update --all (0.0.58)
  - fixing bug to config set/get nested fields, default for wrapper scripts true (0.0.57)
   - change in behavior, instead of `key:value` for set/add `key value` now also works.
  - adding support for container.yaml overrides (0.0.56)
--- a/docs/getting_started/user-guide.rst
+++ b/docs/getting_started/user-guide.rst
@@ -1231,36 +1231,37 @@ it could be that a different method of choosing or sorting tags (beyond the defa
 and filter) is needed.
 
 
-Upgrade
--------
+Sync Registry
+-------------
 
-An upgrade means not just updating tags of existing recipes, but retrieving new
-ones from the main repository. If you clone shpc from the GitHub repository, you can easily achieve
+A sync means not just updating tags of existing recipes, but retrieving new
+ones from the main repository (e.g., syncing with it, or a subset of it). 
+If you clone shpc from the GitHub repository, you can easily achieve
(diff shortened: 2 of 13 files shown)


vsoch:
okay I added a basic test - but I'm thinking I want to refactor so the upgrade command can just be done between the two registries directly.

muffato:
Hi @vsoch . I've taken a quick look at your PR and there's a couple of things I want to ask:
- since `upgrade` only updates the container.yaml, not the installed containers, this is sort-if the opposite of Debian's `apt-cache update` / `apt-get upgrade`. Not sure it's worth keeping the analogy ? I'd suggest something like 'sync_registry' instead
- I wonder why `GitHub` inherits from `Provider`. Is it to be future-proof and support transparent https access without having to change the class hierarchy ? At the moment, since GitHub is accessed through a `git clone`, could you not make `GitHub` inherit from `Filesystem` with some extra init/cleanup functions ?

vsoch:
> since upgrade only updates the container.yaml, not the installed containers, this is sort-if the opposite of Debian's apt-cache update / apt-get upgrade. Not sure it's worth keeping the analogy ? I'd suggest something like 'sync_registry' instead

Yep, definitely can do! Would `sync` alone be too general? (or in other words could there be future uses for sync that would warrant not using  ?)

>  I wonder why GitHub inherits from Provider. Is it to be future-proof and support transparent https access without having to change the class hierarchy ? At the moment, since GitHub is accessed through a git clone, could you not make GitHub inherit from Filesystem with some extra init/cleanup functions ?


I could definitely refactor so Github handles from filesystem! The only main difference is that GitHub needs to clone to populate the source.

vsoch:
okay all set!

- upgrade --> sync-registry
- docs updated
- the sync function is now directly on the registry class
- I added an extra argument to say "don't add new recipes that I don't have"
- GitHub is based off the Filesystem parser. I could get them less redundant for the iter_* functions, however a GitHub remote could have a subdirectory with the registry, and this isn't the case for the filesystem. The Filesystem should always have the root of the registry provided.

marcodelapierre:
Here at Pawsey Centre I have done the following:
1. pip install versioned SHPC
2. git clone and prune SHPC registry from github; still doing it in a versioned way for reproducibility purposes
3. add second, site-specific, version controlled, registry in SHPC settings, to allow add/edit recipes that are not (yet) in the main SHPC repo

That being said, I like the idea of also being able to refer to a remote registry!

vsoch:
hey y'all! So we got sidetracked (at least I did!) with the idea of a remote registry. I think this is something to continue to think about (and maybe pursue) but I don't want to hold up the work here. @surak I think this addition was done to help your use case - should we merge or is there something we need to rethink? I generally was troubled by how hard it is for y'all to get updated recipes, and that's why I was thinking of this remote idea. If you think that's a better approach to take, we can hold off merging here and focus on that. I'd start by making a cloned registry (external to shpc) and then writing interactions with it from shpc, and (maybe?) ultimately removing it here. It does seem like providing it alongside shpc makes it annoying to update, and arguably the entire thing could just be external or remote akin to an OCI registry.

vsoch:
okay I'm jumping in!! Going to start with this branch, and create a new branch that tests nuking the local registry and using this one: https://github.com/singularityhub/shpc-registry

And me of the past was forward thinking - it already has a static API! Lolz.

surak:
Hey, that's great!

One can always have a version of this repo here pointing as a git submodule to that one, right? So this one here is synced anyway.

marcodelapierre:
Great stuff @vsoch! :-) 

One thought: how about version tagging the registry repo, in conjunction with tags in the main SHPC repo? Just as an additional way of achieving reproducibility of deployments.

vsoch:
That's a great idea! How would you go about doing it? Once a month is a new version for the year/month? E.g., `2022.8` ? That would be the easiest way to get the current version I think without looking it up!

vsoch:
Also some good news - I'm almost done with the refactor - I'm going through tests carefully now and making sure we do everything for a local filesystem registry _and_ the remote. I likely won't finish tonight but sooner than later!

marcodelapierre:
I have no strong opinions on the versioning convention, I think what you suggest would work very well! (only.. always 2 digits for the month please :-P )

Issue #570, opened 2022-08-01
Missing features/gpu: true in many nvidia containers


surak:
**Describe the bug**

A number of NVIDIA containers from NVCR.IO simply lack the "features:gpu" tag, so they won't see multi-gpus.

I would expect all NVIDIA-optimized codes to use gpus :-)

For example:
https://github.com/singularityhub/singularity-hpc/blob/main/registry/nvcr.io/nvidia/rapidsai/rapidsai/container.yaml
https://github.com/singularityhub/singularity-hpc/blob/main/registry/nvcr.io/nvidia/hpc-benchmarks/container.yaml
https://github.com/singularityhub/singularity-hpc/blob/main/registry/nvcr.io/nvidia/digits/container.yaml
https://github.com/singularityhub/singularity-hpc/blob/main/registry/nvcr.io/nvidia/tensorflow/container.yaml

vsoch:
This was merged (and I forgot to close) thanks @surak !

PR #571, opened 2022-08-01
Fix for #570

surak:
I haven't checked other containers on the repo, but I think those should be there.


vsoch:
Everything looks good, but the tensorflow container should not be renamed! We will also want to make sure to carry these changes forward to https://github.com/singularityhub/shpc-registry, which will be the replacement for the one here (eventually, have a bit more work to do on it).

surak:
Of course! So, about this duplicity of registries, how about removing the registry completely from here and linking the other one as a git submodule instead? The net effect is that you have two repos and only one registry proper, but fetching this one here gets both.

vsoch:
yes that's the plan :) The default registry will be remote, and the user won't have to think about updating, etc. An interested user (that doesn't want to make web requests) can just clone the registry to get the current behavior (or add as a submodule) - we will have support for a GitHub provider or Filesystem provider!

PR #575, opened 2022-08-01
Make clear that the default won't work with gpus

surak:


surak (commented):
I didn't notice that one has to actively push to have gpu support, as it seemed obvious.

docs/getting_started/user-guide.rst
```diff
@@ -276,7 +276,7 @@ supports gpu":
       gpu: true
      
 Given that you are installing a module for a Singularity container, the ``--nv``
-option will be added. Currently, the following features are supported:
+option will be added IF the container_features is also set - the default is ``null``. Currently, the following features are supported:
```
vsoch, line 279: I would say something like:

> By default, `container_features` in your settings.yaml is unset, meaning that you will not enable any kind of gpu features for containers. You will need to set this feature to something non-null in your settings.yaml to enable features.

surak: Fine for me!

PR #576, opened 2022-08-03
Add/remote registry

vsoch:
This is a _huge_ PR to replace #567 that will make the default registry remote, and add the concept of a registry backend, either remote (GitHub) or a local filesystem. I've done careful work to ensure all tests try both cases, but likely this will need a lot of manual testing to be double sure I didn't break anything!


vsoch:
@muffato @marcodelapierre @surak I pulled it off!

muffato:
Hi @vsoch . That looks very impressive 👏🏼 ! I've started looking at the code, but I haven't run any tests yet. I'll try to get to it this week if that's OK ?

vsoch:
Definitely OK by me!

docs/getting_started/user-guide.rst
```diff
@@ -1228,6 +1260,63 @@ it could be that a different method of choosing or sorting tags (beyond the defa
...
+And finally, you can ask to add new containers and completely update container.yaml files. 
+
+.. code-block:: console
+
+    $ shpc sync-registry --registry ./registry --all
+    
+This means we do a side by side comparison of your filesystem registry and the upstream, and we add new
+recipes folders that you don't have, and we replace any upstream files into recipes that you do have.
+Be careful with this option, as if you've made changes to a container.yaml or associated
+file in the upstream they will be lost. For this reason, we always recommend that you do a dry run first:
```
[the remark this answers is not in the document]
muffato: In a different test, it told me it would add some git object files:
```
Cloning into '/tmp/shpc-tmp.e1ue2gm4'...
remote: Enumerating objects: 68, done.
remote: Counting objects: 100% (68/68), done.
remote: Compressing objects: 100% (37/37), done.
remote: Total 68 (delta 1), reused 52 (delta 1), pack-reused 0
Unpacking objects: 100% (68/68), done.
.git/objects/46 will be added newly.
.git/objects/b8 will be added newly.
.git/objects/6d will be added newly.
.git/objects/5b will be added newly.
.git/objects/40 will be added newly.
.git/objects/19 will be added newly.
.git/objects/28 will be added newly.
.git/objects/44 will be added newly.
.git/objects/83 will be added newly.
.git/objects/30 will be added newly.
.git/objects/38 will be added newly.
.git/objects/98 will be added newly.
```
I'd expect shpc to sync everything but `.git`. If people want the local registry to be a git checkout, _they_ have to deal with it (and do the fetch, commit, etc, themselves)

shpc/main/registry/github.py
```diff
@@ -0,0 +1,174 @@
...
+        Determine if a module exists in the registry.
+        """
+        dirname = self.source
+        if self.subdir:
+            dirname = os.path.join(dirname, self.subdir)
+        return os.path.exists(os.path.join(dirname, name))
+
+    @classmethod
+    def matches(cls, source):
+        return cls.__name__.lower() in source and source.startswith("http")
```
muffato, line 101: This made shpc reject my gitlab-based repository of container.yaml. Do you need specific features from GitHub ?

shpc/main/registry/github.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        Clone the known source URL to a temporary directory
+        """
+        tmpdir = tmpdir or shpc.utils.get_tmpdir()
+
+        cmd = ["git", "clone", "--depth", "1"]
+        if self.tag:
+            cmd += ["-b", self.tag]
+        cmd += [self._url, tmpdir]
+        self.source = tmpdir
```
muffato, line 113: This bit here could work with _any_ git transport mechanism (e.g. incl. SSH) and any git server (e.g. incl. GitLab), which is great ! What do you think it would take to remove the https+github restriction ?

shpc/main/registry/github.py
```diff
@@ -0,0 +1,174 @@
...
+        Update local cache from a registry.
+        """
+        if self._cache and not force:
+            return
+        if not self.source.startswith("http"):
+            self.source = "https://%s" % self.source
+
+        # Check for API
+        org, repo = self.source.split("/")[3:]
+        gh_pages = "https://%s.github.io/%s/library.json" % (org, repo)
```
muffato, line 152: Can you explain this `library.json` ? The one in shpc-registry is (mostly) empty, and has a non-JSON header:
```
shpc-registry (gh-pages=) $ cat pages/library.json 
---
layout: library
permalink: library.json
title: Experiment Factory Library
---
```

muffato (commented):
Looks very good to me. `sync-registry` is a great addition, and I think it can simplify our workflow tremendously. But most importantly, the repository feels so clean with the container.yaml out of the way 😌 !

I've tested a few things (installing and adding a container) and left some comments.

docs/getting_started/user-guide.rst
```diff
@@ -1228,6 +1260,63 @@ it could be that a different method of choosing or sorting tags (beyond the defa
...
+
+    $ shpc sync-registry
+    This command is only supported for a filesystem registry! Add one or use --registry.    
+
+We can then make a dummy directory to support sync. You could also make this directory and add to your settings proper under ``registry``.
+
+.. code-block:: console
+
+    $ mkdir -p ./registry
+    $ shpc sync-registry --registry ./registry
```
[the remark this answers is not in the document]
vsoch: It should technically clone into a temporary location and then copy all the recipes (based on what you have). The clone you see in the console is into a temporary space. Hmm - do you have a suggestion for how to clone and maintain the executable bit? I wouldn't have even noticed that!

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
vsoch: Oh that seems like a bug, because it should not be adding anything aside from the container recipes! I'll put in a fix for this tonight after work.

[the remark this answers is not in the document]
vsoch: Can you give me the command to reproduce this? I want to see if there is a way to only include paths with container.yaml.

docs/getting_started/user-guide.rst
```diff
@@ -1228,6 +1260,63 @@ it could be that a different method of choosing or sorting tags (beyond the defa
...
+    $ shpc sync-registry --registry ./registry --all
+    
+This means we do a side by side comparison of your filesystem registry and the upstream, and we add new
+recipes folders that you don't have, and we replace any upstream files into recipes that you do have.
+Be careful with this option, as if you've made changes to a container.yaml or associated
+file in the upstream they will be lost. For this reason, we always recommend that you do a dry run first:
+
+.. code-block:: console
+
+    $ shpc sync-registry --registry ./registry --dry-run
```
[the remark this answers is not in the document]
vsoch: Sure! We can do that for commands that have consequences we don't want.

docs/getting_started/user-guide.rst
```diff
@@ -1228,6 +1260,63 @@ it could be that a different method of choosing or sorting tags (beyond the defa
...
+-------------
+
+A sync is when we take your local filesystem registry, and retrieve updates from the remote defined at 
+``sync_registry`` in your settings.yaml. Since sync will be writing recipes to the filesystem
+it only works if you target a filesystem registry (meaning that the default registry
+as a remote will not work). 
+
+.. note::
+
+   By default, the first filesystem registry found in your settings under the registry list will
```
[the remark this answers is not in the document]
vsoch: Hmm, I like this idea, but I don't think it belongs in the settings.yaml - the reason because there is a distinction between defining your registries and then asking for sync. What about a yaml file that you provide to sync instead?

shpc/main/registry/github.py
vsoch: Nope, I think we could support both, it's just a clone.

vsoch: Ah - so I take it back - there is a GitHub pages component. If we can get a working version to serve the same kind of Github pages (does GitLab have that?) I can develop this backend. See under main/registry/github.py the function called _update_cache.

shpc/main/registry/github.py
vsoch: The ideal bit about GitHub is that it has GitHub pages to provide a static "API" that makes the interactions more efficient (e.g., I can get contents without clone). My suggestion is to start with GitLab and put up a registry with this same automation and pages, and then I can write the controller. Then we can tackle the others, and we just won't have this ability to look before clone. Would that work?

shpc/main/registry/github.py
vsoch: Take a look in `_layouts/library.json` and it will make sense. This provide a static API (RESTful) from the repository that we can get and cache from GitHub pages. And the Experiment Factory is a typo - that's from where I've made one like this before. :)

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
muffato: If we're talking about adding a yaml file, what about defining a standard one to put at the root of the local registry, and that sync would automatically use ? Similar in a way to view.yaml at the root of each view.

The file could initially just contain the URL of the remote, but could be expanded. First thing that comes to my mind would be to allow the local registry to be in sync with a sub-directory of the remote, e.g.
```
url: https://github.com/remote
path: my_registry
```
the remote git repository would have a directory named `my_registry` at the root, and only it would be synchronised.

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
muffato: "copy all the recipes" ? Not just the recipes but other files too, isn't it ? Otherwise these two scripts couldn't have been modified

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
vsoch: @muffato I don't think we'd need this by default - the main shpc can work entirely without any registry (just the remote in the config) and certainly a remote doesn't need the reference. I think we should provide examples that people can copy, but it shouldn't be some default or required thing. To summarize the use cases:

1. I'm using shpc installing from a remote registry, and I don't care to add or sync. The default works without extra files. (basic)
2. I'm using shpc install from a remote and wanting to sync recipes locally to tweak. I can just add the local registry to my settings.yaml (intermediate)
3. I have several remote registries I want to always sync from, and I'll add an extra file to specify this design (advanced).

Others can chime in, but I think 1 and 2 are the common use cases, and 3 is less common, and given this, I don't want to have it be a requirement for a registry, but rather something that is easy to add if you want that advanced functionality. And it would look like:

```
registry: ["/path/to/local1", "https://remote1", "/path/to/local2", "/path/to/local3"]
sync_registry:
    "/path/to/local1": "https://sync_remote1"
    "/path/to/local3": "https://sync_remote3"
```
as you specified.  Thoughts?

shpc/main/registry/github.py
vsoch: Okay typo above is fixed!

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
vsoch: This is added in https://github.com/singularityhub/singularity-hpc/pull/576/commits/a55c243cf55c7667916b11a24c62322725c77593

vsoch:
@muffato I've done various tweaks (and ran isort/black) https://github.com/singularityhub/singularity-hpc/pull/576/commits/a55c243cf55c7667916b11a24c62322725c77593.

For next steps:

- I want to add GitLab but we will need a template first. I'm going to make a new issue for this that I will tackle after the initial work here.
- I want to add functionality for the registry file - can you give me the example you were wanting to use (or do we need another remote to be made first?)
- Can you give me the sync commands that borked your registry? I am going to see if I can add more logic to only consider container modules (and a test for it).

vsoch:
okay - and got started on creating a GitLab equivalent! https://gitlab.com/singularityhub/shpc-registry supposedly pages take 30 minutes to deploy... so not sure if I did it right yet! I also realized default branches there are "master" so I added a branch arg to the docgen command here.

vsoch:
okay Gitlab registry (and remote here) is added, and I un-broke the CI! :laughing:

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
vsoch: okay this is added (pending passing tests). The first bit seems redundant if we just need the key value pairs, so I only added `sync_registry`

vsoch:
@muffato I've finished 2/3 things we discussed above - creating a GitLab remote (and support for it) and configuring sync from file. The last bullet I'll need some pointers from you for how to reproduce the bugs you saw (and to be sure we test for them!)

shpc/main/registry/__init__.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    if not os.path.exists(existing_path):
+        shpc.utils.mkdir_p(existing_path)
+    for filename in shpc.utils.recursive_find(from_path):
+        relative_path = filename.replace(from_path, "").strip("/")
+        to_path = os.path.join(existing_path, relative_path)
+        if os.path.exists(to_path):
+            shutil.rmtree(to_path)
+        shpc.utils.mkdir_p(os.path.dirname(to_path))
+        shutil.copyfile(filename, to_path)
```
muffato, line 32: This is where the copy happens. According to the docs, this only copies the content of the file, not its mode or other stats. That's why it dropped the executable bits from `vanessa/salad/*_fork.sh`.
The simplest solution (tested) is to use the full-blown `copy`, which copies everything:
```suggestion
        shutil.copy(filename, to_path)
```

vsoch:
Okay - I opted for https://docs.python.org/3/library/shutil.html#shutil.copy2 because that also preserves metadata (in addition to data and permissions).

muffato (commented):
Hi. FYI I'm looking at the code that does the copying, to make it exclude `.git`. I found where the bug is, and I'll submit a PR onto this PR later today

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
muffato: On the freshest `add/remote-registry` with no local modifications:
```
$ mkdir test_sync
$ shpc config add registry $PWD/test_sync
$ shpc sync-registry --dry-run
(...)
adminer will be added newly.
.github will be added newly.
.github will be added newly.
.github will be added newly.
.github will be added newly.
.github/workflows will be added newly.
.github/workflows will be added newly.
.github/workflows will be added newly.
.github/ISSUE_TEMPLATE will be added newly.
.github/ISSUE_TEMPLATE will be added newly.
.github/ISSUE_TEMPLATE will be added newly.
.github/ISSUE_TEMPLATE will be added newly.
.git will be added newly.
.git will be added newly.
.git will be added newly.
.git will be added newly.
.git will be added newly.
.git will be added newly.
.git/logs will be added newly.
.git/logs/refs/heads will be added newly.
.git/logs/refs/remotes/origin will be added newly.
.git/objects/pack will be added newly.
.git/objects/pack will be added newly.
.git/refs/remotes/origin will be added newly.
.git/refs/heads will be added newly.
.git/info will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
.git/hooks will be added newly.
$ shpc sync-registry
(...)
adminer will be added newly.
.github will be added newly.
.git will be added newly.
```

The perfect solution would be parsing all the container.yaml, extracting the templates they refer too, but also parsing those templates in case they include other files, etc, and also deal with the per-tag yaml file that's been recently introduced. I guess it's technically doable, but the code may be quite complex :)
I would simply suggest to copy everything but the `.git` directory, or everything but the hidden files+directories.

docs/getting_started/user-guide.rst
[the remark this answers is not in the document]
muffato: I can't reproduce this one precisely. But I've got something similar. On a clean checkout with no local modifications
```
$ rm -rf test_sync/
$ git clone git@github.com:singularityhub/singularity-hpc.git test_sync
$ cd test_sync/
$ git remote remove origin
$ cd ..
$ shpc config add registry $PWD/test_sync
$ shpc sync-registry
(...)
adminer will be added newly.
.git/logs/refs/remotes/origin will be added newly.
.git/refs/remotes/origin will be added newly.
```

muffato:
I'm not sure where I'm going with my changes, so best is that I describe what I've noticed.

- `sync_from_remote` identifies what needs to be copied by first calling `iter_modules` (it then calls `update_container_module` when required).
- `iter_modules` calls `shpc.utils.recursive_find` and then discards the _files_ that start with a dot, therefore allowing the _directories_ that start with a dot.

vsoch:
@muffato give that a shot! I'm going for a run, then probably a break/dinner, I promise I'll be back later tonight after that! :avocado:

muffato:
https://github.com/singularityhub/singularity-hpc/pull/576/commits/8739676d499f77faf196d76a5bb9d1601121da31 made the trick, fantastic !

vsoch:
Sounds good!

shpc/client/sync.py
```diff
@@ -0,0 +1,31 @@
+__author__ = "Vanessa Sochat"
+__copyright__ = "Copyright 2021-2022, Vanessa Sochat"
+__license__ = "MPL 2.0"
+
```
muffato, line 4: For `os.path.exists` below

```suggestion

import os

```

muffato (commented):
Hi

The yaml file works great (with that little addition) ! And I've also tested the check for unknown arguments on the command-line: a life saver !

About GitLab:
 - `shpc show` breaks if the repo is under groups because shpc does `org, repo = self.source.split("/")[3:]`, while gitlab allows groups in between the org and repo name, e.g. `https://gitlab.internal.sanger.ac.uk/tol-it/software/shpc-installation.git`
 - Once moved to match `org, repo`, then it fails because I don't have anything under `https://$org.gitlab.io/$repo/library.json`, and it's due to two things:
   - I haven't set up `.gitlab-ci.yml` (I guess that's what's needed to generate `library.json` ?)
   - Even if had it, our "pages" are at `https://$org.pages.internal.sanger.ac.uk/$repo_name` (I don't know how it looks like with groups !)
 - At least `shpc sync-registry` works, even with our internal one

Questions:
- How important is this `library.json` API ? In case it can't be found, what about taking a temporary local checkout and building a cache from there ?
- Then, what about ssh support ? If a substitute for `library.json` can be found, I guess we don't require https any more, and ssh would be just fine (since all git access would be done via `git clone`)
- How / where to define the tag/branch and the subdir ? I can see the code supports that, but only the tag option in `sync-registry`
- Do you really want to support on-prem installations of gitlab ? and do they all follow this pattern ?

vsoch:
@muffato try using my dummy as an example: https://gitlab.com/singularityhub/shpc-registry. That should have the workflow to generate and deploy to pages - I'm kind of a GitLab noob but after that I was able to click pages and make sure my artifact in "public" worked. For the other stuffs I'll take a closer look and make some tweaks tonight, and in the meantime if you want to share the update to match "org" and "repo" I can be sure to have it handy later! We should probably have a fallback for show that checks for the library and then issues an error if it doesn't exist, and tells the user to visit the repository if they want to browse. More specifically:

>    How important is this library.json API ? In case it can't be found, what about taking a temporary local checkout and building a cache from there ?

I think this is important for quick query - e.g., `shpc show`. I wouldn't want it to do a clone to tmp every time someone ran that! But as I mentioned above, I think we can add a fallback that tells the user to visit the repository if there is no library.json. The alternative (if it's available without token auth) is to use some GItHub/GitLab API to list the top level contents of the repo (in the same way that we do after we clone).

> Then, what about ssh support ? If a substitute for library.json can be found, I guess we don't require https any more, and ssh would be just fine (since all git access would be done via git clone)

We could definitely look into this! My thinking is we should work on this after this PR (which is already very big). If you want to open an issue with some details about how this works / what you want to work we can have it handy for when that time comes.

> How / where to define the tag/branch and the subdir ? I can see the code supports that, but only the tag option in sync-registry

Currently it's just via the command line, so you can provide it with or without a file. I think we might honor the branch but I don't recall if we pass forward a subdirectory preference, the reason being we assume a remote registry is going to honor a convention of always having recipes at the root (do you see a use case where this isn't the case?).

>  Do you really want to support on-prem installations of gitlab ? and do they all follow this pattern ?

I don't know - what do you think? My senses are we should support official Gitlab, and if on-prem isn't too hard, we should strive for that too.

muffato:
Status update:
- I could get something to run our GitLab CI, but it's not working properly: it can't find any of the local recipes
- I've found the URL format of GitLab Pages when there are groups or sub-groups, and can make a PR for that
- Re. on prem, GitLab Pages can actually be deployed anywhere, it's left to the administrators to decide. shpc can't find this out by itself without a dedicated config option



> the reason being we assume a remote registry is going to honor a convention of always having recipes at the root (do you see a use case where this isn't the case?).

The other shpc asset is the general template scripts (i.e. ones that can be shared between recipes and therefore are not in the same directory as `container.yaml`). Having both the registry and the templates in the same repository can be useful to track things under the same git history

muffato:
I got the CI to work ! It was doing a `pip install` of the `main` branch. Changed it to `add/remote-registry`

shpc/main/registry/remote.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Check for API
+        org, repo = self.source.split("/")[3:]
+        gh_pages = "https://%s.%s.io/%s/library.json" % (org, self.provider_name, repo)
+        response = requests.get(gh_pages)
+        if response.status_code != 200:
+            sys.exit(
+                "Remote %s is not deploying a Registry API. Open a GitHub issue to ask for help."
+                % self.source
+            )
```
muffato, line 158:
```suggestion
        url = self.source
        if not url.startswith("http"):
           url = "https://%s" % url
        if url.endswith(".git"):
            url = source[:-4]

        # Check for API
        url_parts = url.split("/")[3:]
        gh_pages = "https://%s.%s.io/%s/library.json" % (url_parts[0], self.provider_name, "/".join(url_parts[1:]))
        response = requests.get(gh_pages)
        if response.status_code != 200:
            sys.exit(
                "Remote %s is not deploying a Registry API (%s). Open a GitHub issue to ask for help."
                % (self.source, gh_pages)
            )
```

muffato (approved):
I think that's all I wanted to test. I'll create issues with additional features that can be addressed later

shpc/main/registry/remote.py
vsoch: Can you walk me through this change? Why wouldn't we want to change the source (for good) with a proper https url?

muffato:
FYI I will _not_ create an issue to support cloning over ssh, as the only advantage I think it'd bring is being able to clone private repositories. But shpc anyway requires a *public* library.json, which defeats the purpose of keeping the registry private.

vsoch:
@muffato I think we could have support for that still, it just would be a different kind of clone (and show/list would do a directory listing instead of library.json).

muffato:
🤔 . You also said
> I think this is important for quick query - e.g., shpc show. I wouldn't want it to do a clone to tmp every time someone ran that! 

I'm confused 😅 !

vsoch:
There are two different cases here:

- Anything with a public GitHub (or GitLab) url should work with show. If pages is available, we expect a library.json.
- Anything that is private we cannot enforce that expectation (ssh).

shpc/main/registry/remote.py
muffato: About this particular change, I don't understand why the source would not start with `https://` at this stage, since the constructor checks that and raises an error.
That's why I had suggested to remove it in https://github.com/singularityhub/singularity-hpc/pull/579/files#diff-183e41e47a9fe06125f88dd47d33eb5bc399a6458ad3bc0e8d5c98e422179239L147-L148

vsoch: I did that because a person may just want to specify it as `github.com/<org>/<name>` . I don't see that the constructor checks and throws an error, it seems to just set it: https://github.com/singularityhub/singularity-hpc/blob/132d04ad4e41abaafd96c762961f48735e5de682/shpc/main/registry/remote.py#L86.

muffato: The check happens in the super-class https://github.com/singularityhub/singularity-hpc/blob/f948e4ecffc43dc595bdc7e6b62298399f6a02ef/shpc/main/registry/provider.py#L35-L40

vsoch: But there is an edge case, if they are in a go src directory and there is `github.com` it will pass (and then later fail).

vsoch: This is a more of a "quick bail out before we do anything" check.

vsoch: But not to get distracted with this check, which could be argued is redundant - can you explain the changes in the above? What problem is it fixing? I think we are close and (things look generally ok) and could merge/release soon.

muffato: The patch is to allow support for "groups" in gitlab, cf your message (https://github.com/singularityhub/singularity-hpc/pull/576#issuecomment-1221642688):
> and in the meantime if you want to share the update to match "org" and "repo"

muffato: with groups, the url is https://gitlab.com/group/sub-group/repo.git

vsoch: Ah gotcha, thank you for clarifying! Let me see if I can organize the functions a bit - brb!

vsoch: okay check out https://github.com/singularityhub/singularity-hpc/pull/576/commits/1f29489067e6aa639b214a829bcd57aabf3f62ca. !

vsoch:
@muffato let me know if you have more questions for the review! Apologies my responses are a bit scattered.

muffato:
I get from your previous message that you don't see the registry providing a pseudo API and `library.json` as a hard requirement anymore. So, I can create a ticket about adding support for private repositories (access could be done over https and ssh), and we'll deal with the implications of not having `library.json` there ?
Apart from that, I think I had tested everything I wanted to test, and I should be able to approve this PR. I'll double-check later today if that's OK ?

vsoch:
> I get from your previous message that you don't see the registry providing a pseudo API and library.json as a hard requirement anymore. So, I can create a ticket about adding support for private repositories (access could be done over https and ssh), and we'll deal with the implications of not having library.json there ?

Yes absolutely, I do think we should support this - encourage folks to provide the extra metadata but it shouldn't be a hard requirement. Open away!

> Apart from that, I think I had tested everything I wanted to test, and I should be able to approve this PR. I'll double-check later today if that's OK ?
Yes totally! I'm in a workday I won't be able to do another pass and/or release until the evening. Thank you @muffato !

docs/getting_started/user-guide.rst
```diff
@@ -364,10 +367,24 @@ Registry
...
+2. Clone the default remote to a local filesystem folder and manage manually (e.g., git pull)
+3. Create your own local registry in addition (or without) the remote.
+4. For any local registry, either sync (``shpc sync``) or add a recipe locally from a remote (``registry-add``).
+
+If you want to do the first, no further action is needed! Each of these remaining
+examples will be described here.
+
+Use the default 
+
+TODO WRITE THESE OUT SHOW HOW TO UPDATE
```
muffato, line 387: TODO

muffato (commented):
Can you add to the developer guide instructions for making your own registry ? (especially setting up the CI)

docs/getting_started/user-guide.rst
vsoch: lol, I think I just totally forget - yes let me add a few lines!

vsoch: See https://github.com/singularityhub/singularity-hpc/pull/576/commits/da4c91cee4fd95c8bbfed314e0e4f909069c1524 !

muffato (approved):

muffato:
https://user-images.githubusercontent.com/623458/187009228-3b4e2399-9556-4bec-b0cf-5b927e224c63.mp4

PR #577, opened 2022-08-12
Upgraded Rust (and moved to the official image)

muffato:
The autamus package hasn't been updated in a while, and I've found some code that requires a recent version of Rust. Therefore decided to move to the official Rust image.
I had to empty `CARGO_HOME` (couldn't find how to _unset_ it) because otherwise `cargo build` wants to store all the files in the container's (read-only) filesystem


vsoch:
I’m happy to add the new recipe, but for now my preference would be not to delete the previous autamus. Also @muffato if we change to the remote registry (which I’ve already cloned) you’d need to re-add there. Have you taken a look at the PR yet?

muffato:
Yes, of course, let's not merge this PR, and I'll open one on the other side once the move to a remote registry is enacted.

By the way, why would you not remove the autamus one ? If it's to be able to cater for older rust releases, I can extend the container.yaml and add those older versions back.

vsoch:
> By the way, why would you not remove the autamus one ? If it's to be able to cater for older rust releases, I can extend the container.yaml and add those older versions back.

It's mostly not removing a recipe that people might be using - autamus development is paused, but I think @alecbcs has plans to continue it at some point, and I don't want to start purging the recipes from shpc! And definitely - if there is a particular container that is useful we could do some manual updates.

PR #579, opened 2022-08-21
Misc. refactoring

muffato:
Hi @vsoch 
I went down the rabbit hole refactoring things here and there., and I didn't even get to fixing how the hidden files / `.git` were handled :/ But here is what I have

In short, I was trying to move things from `FileSystem` and `VersionControl` to `Provider`, thinking it would be easier to solve issues. I did that by making `self.source` include the subdir in `VersionControl` so that all instances of providers can directly rely on `self.source`.

I also try to deduplicate the checks that both `Provider`'s constructor and the `matches` methods were doing

Finally I made `Provider` know less about `GitHub` and `GitLab`

### DIFF
--- a/shpc/main/registry/__init__.py
+++ b/shpc/main/registry/__init__.py
@@ -84,7 +84,7 @@ class Registry:
             for registry, module in reg.iter_modules():
                 yield registry, module
 
-    def get_registry(self, source):
+    def get_registry(self, source, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


vsoch:
Hey @muffato ! Thank you for taking a look! Did you see that I pushed a fix (did you get to try it out?) and was there still an issue?

PR #589, opened 2022-09-04
Support accessing remote registries via ssh

muffato:
(extension of #585 )
Note: the URL has to be given as `ssh://[user@]host.xz[:port]/path/to/repo.git` rather than the shorter version `[user@]host.xz:path/to/repo.git`.

Comments: 
1. The shpc code was sometimes using `https://` to test the remote URLs, sometimes `http`, and I've kept the same dichotomy, i.e. `ssh://` and `ssh`. Any particular reason why ? I'd prefer only seeing `https://` (and `ssh://`)
2. My change is essentially piggy-backing on the existing `Provider` class, which can only be instantiated as `GitHub` or `GitLab`. Since there is now a fallback for when Pages are not available, remote registries can be anything else that these two forges. Furthermore, the transport protocol can be anything that git supports, incl. `git://` and even `ftp://` ! Remote registries could perhaps be recognised as the ones that contain `://` (and are not valid local paths) ?

Happy to do that refactoring if you like !


muffato:
(I'm aware that `black` doesn't like the long lines. I'll fix these later)

vsoch:
> The shpc code was sometimes using https:// to test the remote URLs, sometimes http, and I've kept the same dichotomy, i.e. ssh:// and ssh. Any particular reason why ? I'd prefer only seeing https:// (and ssh://)

At first I was just checking for http to be able to hit http or https, and I agree it would make sense to require https, however then we'd have to do an updated check to also check for http, and update to https if provided (with a warning). The second check I believe I had originally set the https prefix and assumed it would be there, but with all the changes this might be wrong now.

> My change is essentially piggy-backing on the existing Provider class, which can only be instantiated as GitHub or GitLab. Since there is now a fallback for when Pages are not available, remote registries can be anything else that these two forges. 

Nice!! I really appreciate you taking initiative on adding ssh - it would have been my next TODO but there is a lot in the queue so I'm a bit slow!

> Furthermore, the transport protocol can be anything that git supports, incl. git:// and even ftp:// ! Remote registries could perhaps be recognised as the ones that contain :// (and are not valid local paths) ?

Indeed the transport protocol could be any of those, and perhaps we should be flexible to that.

shpc/main/registry/provider.py
```diff
@@ -33,7 +33,7 @@ class Provider:
     """
 
     def __init__(self, source, *args, **kwargs):
-        if not (source.startswith("https://") or os.path.exists(source)):
+        if not (source.startswith("https://") or source.startswith("ssh://") or os.path.exists(source)):
```
vsoch, line 36: I agree with you here this could be more elegant - what if we have more of a regular expression that is assembled to match any of those potential protocols and then properly does the check/replacement?

shpc/main/registry/remote.py
```diff
@@ -190,6 +190,8 @@ def _update_cache(self, force=False):
         if self._cache and not force:
             return
 
+        if self.source.startswith("ssh"):
```
vsoch, line 193: We could do the same here, e.g., `if re.search("<pattern for acceptable protocols">, self.source):`

muffato:
New version coming in with a lot of changes, incl. some I rescued from #579  !

- Currently I allow any remote as long as it contains `://`. That could easily be changed to the regex as you were suggesting because ...
- ... the numerous tests for `http`, `https`, `os.path.exists`, etc, have been consolidated into just two locations: `Filesystem` for local paths, and `VersionControl` for remote paths.
- I added config options in `VersionControl` to define the URLs of library.json and container.yaml based on the domain name (i.e. github.com and gitlab.com)
- I tidied up the constructors and accessors of `Filesystem` and `VersionControl`. In particular, `VersionControl` has `url` for the URL, and `source` for the clone on disk, which allows moving `iter_modules` to the parent class

shpc/main/registry/remote.py
```diff
@@ -91,44 +74,69 @@ def override_exists(self, tag):
...
+
+    def __init__(self, source, tag=None, subdir=None):
+        if "://" not in source:
+            if os.path.exists(source):
+                raise ValueError(
+                    "VersionControl registry must be a remote path, not a local one."
+                )
+            # Normalise the URL
+            # Heuristics: if there is a @, it's probably ssh
+            if "@" in self.url:
```
vsoch, line 104: Just making a suggestion to get the tests passing, and then I can do a more thorough review.

```suggestion
            if "@" in source:
```
For this loop, we are doing a combination of checking source and then setting the self.url. This leads to the double else, which is a bit too nested for my preference. I think I'd like better to set self.url = source off the bat, and then just have one if statement. If we don't hit the if statement, the self.url remains the same.

I haven't reviewed closely yet, but I see what you are trying to do and I like it!

shpc/main/registry/remote.py
```diff
@@ -91,44 +74,69 @@ def override_exists(self, tag):
...
         # E.g., subdirectory with registry files
-        self.subdir = kwargs.get("subdir")
-        super().__init__(*args, **kwargs)
-        self._url = self.source
+        self.subdir = subdir
 
     @classmethod
     def matches(cls, source):
-        return cls.provider_name in source and source.startswith("http")
+        return ("://" in source) or not os.path.exists(source)
```
vsoch, line 123: Any reason for the extra parens? they shouldn't be needed here - it's one or with two checks, so there is only one option for the order.

shpc/main/registry/remote.py
```diff
@@ -91,44 +74,69 @@ def override_exists(self, tag):
...
 
     @property
-    def source_url(self):
+    def library_url(self):
```
vsoch, line 126: Good thinking with this name, indeed it is the library url!

shpc/main/registry/remote.py
```diff
@@ -138,48 +146,35 @@ def exists(self, name):
         if self._cache and name in self._cache:
             return True
         dirname = self.source
-        if self.subdir:
```
vsoch, line 141: Hmm Do we have a test case that triggers this? Do we set a default for self.source? It could be possible that it's either unset or set to something unexpected. Should the behavior differ based on is_cloned?

shpc/main/registry/remote.py
```diff
@@ -190,8 +185,11 @@ def _update_cache(self, force=False):
         if self._cache and not force:
             return
 
+        library_url = self.library_url
```
vsoch, line 188: If this is a property my preference is to reference it directly.

shpc/main/registry/remote.py
```diff
@@ -138,48 +146,35 @@ def exists(self, name):
...
-                continue
-            yield dirname, module
-
     def find(self, name):
         """
         Find a particular entry in a registry
         """
-        if not self._cache:
-            self._update_cache()
+        self._update_cache()
```
vsoch, line 177: Good call - this checks if it's already there and only does it when force=true.

shpc/main/registry/remote.py
```diff
@@ -216,6 +214,21 @@ def _update_clone_cache(self):
...
 
+    def get_module_config_url(self, module_name):
+        """
+        Get the raw address of the config (container.yaml)
+        """
+        parts = self.url.split("/")
+        domain = parts[2]
+        if domain in self.raw_container_url_schemes:
+            t = ("/".join(parts[3:]), self.tag, module_name)
+            return (
```
vsoch, line 225: if you have a return like this for an if, you don't need the else - it will hit the else condition by not returning (just a design preference!)

shpc/main/settings.py
```diff
@@ -241,32 +241,6 @@ def change_validate(self, key, value):
...
-    def filesystem_registry(self):
-        """
-        Return the first found filesystem registry
-        """
-        for path in self.registry:
-            if path.startswith("http") or not os.path.exists(path):
-                continue
-            return path
-
-    def ensure_filesystem_registry(self):
```
vsoch, line 254: I don't remember all the places these were used = but it would be good to do a sanity check grep to ensure it's not needed anymore (I'm away from my main computer now).

vsoch (commented):
And possibly would be good to add an example of each registry type (ssh, git, https, etc) to the docs - I forget these formats all the time!

shpc/main/registry/remote.py
muffato: That's my style :) I add parentheses to help my brain identify the logical blocks, even if they're not needed due to the precedence of the operators. I actually had this one as `return (("://" in source) or not os.path.exists(source))` initially but remove the outer set of parentheses because of `black` ;)

shpc/main/settings.py
muffato: I used `grep` to find the places it was used, and corrected those.

shpc/main/registry/remote.py
muffato: Originally, self.source was first set to the URL, and then turned into the directory of the clone.
With my change, it remains unset until the repository is cloned, and then it's the directory + the sub-directory.

No there isn't a test case for that

shpc/main/registry/remote.py
muffato: Yes it's a property, but the value needs to be used there and a couple of lines below. Since properties are not cached by default, I  thought I'd essentially make a local cache by keeping the value in the local variable `library_url`

shpc/main/registry/remote.py
muffato: Changed

shpc/main/registry/remote.py
muffato: Changed

shpc/main/registry/remote.py
muffato: Changed

shpc/main/registry/remote.py
vsoch: okay, let's add one more check here then because if for some reason the cache is empty and we haven't cloned yet, it's going to try to os.path.join a None and throw an error. It's up to you what you want to do default to - either it could a warning, or a clone right then, etc,

shpc/main/registry/remote.py
vsoch: Gotcha. So let's instead add self._library_url to do that caching, so when you call the property it checks for that first and returns it if it's defined, otherwise sets it and then returns it. Properties should be able to be used like that, and without copying them as we see here (it's not expected).

vsoch (commented):
Loving these changes! :partying_face: I made a few final comments and I think after those we should be ready for merge,

shpc/main/registry/remote.py
muffato: Changed and pushed (together with the other changes I did earlier today and forgot to push),

shpc/main/registry/remote.py
muffato: Looking at where `source` is called throughout the codebase, I can indeed see some weaknesses. I'll fix these. Stay tune :)

shpc/main/registry/remote.py
```diff
@@ -216,6 +217,20 @@ def _update_clone_cache(self):
...
         shutil.rmtree(tmpdir)
 
+    def get_module_config_url(self, module_name):
+        """
+        Get the raw address of the config (container.yaml)
+        """
+        parts = self.url.split("/")
+        domain = parts[2]
+        if domain in self.raw_container_url_schemes:
+            t = ("/".join(parts[3:]), self.tag, module_name)
```
muffato, line 227: @vsoch . If I do `parts = self.url.split("/", 3)` above, then this becomes `t = (parts[3], self.tag, module_name)`. Do you have a preference ?

vsoch: That seems elegant! The current way is good, but I like that one too :)

muffato:
Major change coming ! Overall, I've tried to straighten out the use of `registry.source` (which only exists in `Filesystem`, not `VersionControl`), access to the underlying clone of remote registries. All methods in `VersionControl` now properly deal with the URL / local clone, and build / use the cache. The clone is only used in two places, and is clearly a transient construct.
Importantly, `shpc.utils.recursive_find` now returns paths *relative* to the start directory. This is because _every single_ caller was doing that.

In other news:
- `update all` can directly update each module entry, rather than going through module names and having to search them in the registry
- `Provider.iter_modules` yields module names only, i.e. without `self.source`, since the source (or URL) is static and publicly accessible
- `Registry.iter_modules` now yields the instance of `Provider` the module is in
- `Registry.exists` returns the instance of `Provider` that matched rather than a path on disk, since such paths don't make sense for remote providers

vsoch:
woot! I'll review this in detail tomorrow (later today) - still early morning and I'm going to go for a little more sleeps.

shpc/main/client.py
```diff
@@ -124,13 +124,12 @@ def update(self, name=None, dryrun=False, filters=None):
         """
         # No name provided == "update all"
         if name:
-            modules = [name]
-        else:
-            modules = [x[1] for x in list(self.registry.iter_modules())]
-
-        for module_name in modules:
-            config = self._load_container(module_name)
+            config = self._load_container(name)
```
vsoch, line 127: I think generally it's good to be consistent with respect to how we handle the list vs. all - e.g., we prepare a list of names for both cases and then call the same functions. With the changes here, it's technically done two different ways, and that could eventually lead to bugs.

shpc/main/registry/__init__.py
```diff
@@ -44,21 +43,29 @@ def __init__(self, settings=None):
...
+                return reg
 
     def iter_registry(self, filter_string=None):
         """
         Iterate over all known registries defined in settings.
         """
         for reg in self.registries:
-            for entry in reg.iter_registry(filter_string=filter_string):
-                yield entry
+            yield from reg.iter_registry(filter_string=filter_string)
```
vsoch, line 68: Oh this is neat, I haven't used "yield from" before, it's a nicer way to do this.

shpc/main/registry/__init__.py
```diff
@@ -152,7 +157,6 @@ def _sync(
             add_new=add_new,
             local=local,
         )
-        remote.cleanup()
```
vsoch, line 155: Any reason that you removed this? If it's a temporary clone I think we'd want to cleanup.

shpc/main/registry/__init__.py
```diff
@@ -171,18 +175,19 @@ def sync_from_remote(
 
         # No local registry provided, use default
         if not local:
-            local = Filesystem(self.settings.filesystem_registry)
+            local = self.filesystem_registry
 
-        tmpdir = remote.source
-        if tmpdir.startswith("http") or not os.path.exists(tmpdir):
-            tmpdir = remote.clone()
+        ini_remote = None
```
vsoch, line 180: What does ini_remote mean?

shpc/main/registry/__init__.py
vsoch: Ah I see you moved it into the sync function! I think my preference is to have it up here - the reason is because it allows for using the underlying function without doing the cleanup. The cleanup is on the wrapper that the client exposes, and shouldn't require a developer user to have it done every time.

shpc/main/registry/filesystem.py
```diff
@@ -99,25 +110,13 @@ def find(self, name):
         if os.path.exists(container_yaml):
             return FilesystemResult(name, container_yaml)
 
-    def cleanup(self):
```
vsoch, line 102: So this was moved elsewhere? Someone (e.g., a developer user) might still want this ability.

shpc/main/registry/provider.py
```diff
@@ -32,24 +34,6 @@ class Provider:
...
     """
 
-    def __init__(self, source, *args, **kwargs):
-        if not (source.startswith("https://") or os.path.exists(source)):
-            raise ValueError(
-                "Registry source must exist on the filesystem or be given as https://."
-            )
-        self.source = source
-
-    def exists(self, name):
```
vsoch, line 42: What do you see as being the intended future use for this class? It seems so abstract and slimmed down I'm not sure it provides anything useful.

shpc/main/registry/provider.py
```diff
@@ -32,24 +34,6 @@ class Provider:
     A general provider should retrieve and provide registry files.
     """
 
-    def __init__(self, source, *args, **kwargs):
```
vsoch, line 35: So this is just a very abstract class at the moment.

shpc/main/registry/remote.py
```diff
@@ -202,19 +211,32 @@ def _update_clone_cache(self):
...
                 "config_url": config_url,
             }
-        shutil.rmtree(tmpdir)
+        self.cleanup()
+
+    def get_module_config_url(self, module_name):
+        """
+        Get the raw address of the config (container.yaml)
+        """
+        (_, _, domain, repo_path) = self.url.split("/", 3)
```
vsoch, line 232: Generally my preference is not the tuplized version, so e.g., without the extra () - there were a few instances of this added here!

vsoch:
All of the above comments make sense! My only comment is for:

> Registry.exists returns the instance of Provider that matched rather than a path on disk, since such paths don't make sense for remote providers

If we clone a remote to the filesystem it would have a path on disk.

shpc/main/registry/filesystem.py
muffato: Entirely delete a local registry ? really ? I removed it because I couldn't understand the use-case, but I can drop the commit if needed

shpc/main/registry/__init__.py
muffato: I moved the `cleanup()` to the same place as the `clone()`, so that precisely `sync_from_remote` cleans after itself, and the developer users don't have to remember that they need to call `cleanup()` themselves every time they call `sync_from_remote`. Especially as there is currently no need for the clone to persist beyond `sync_from_remote`.

shpc/main/client.py
muffato: I improved it the best I can. Otherwise, we need to:

- either inline `_load_container` and `load_registry_config`, duplicating their logic
- or revert to the original inefficient search (ask all registries for their module names and then, for each module name, ask each registry if it knows about it)

shpc/main/registry/__init__.py
muffato: `ini_remote` is the original registry, an instance of `VersionControl`, while `remote` is replaced with a local registry object (`Filesystem`) from the clone.

shpc/main/registry/provider.py
muffato: I couldn't find anything in common between the implementations of the two sub-classes. There could still be a base class, but I think it should list all the methods that are expected from the sub-classes, and throw a "Not implemented" error, rather than doing `pass`

muffato:
> If we clone a remote to the filesystem it would have a path on disk.

But then it essentially becomes a local registry, which is exactly what I do in https://github.com/singularityhub/singularity-hpc/blob/2423714e00e3ef0cc2e7fac65ee483ae08107554/shpc/main/registry/__init__.py#L183

shpc/main/registry/filesystem.py
vsoch: I think the interface between providers should be consistent, yes. If someone clones a remote to a temporary directory to do some work, instantiating it as a Filesystem registry, after they should be able to clean it up.

shpc/main/registry/provider.py
vsoch: Agree - that's what I typically do.

shpc/main/registry/__init__.py
vsoch: ah so ini == initial? or init? I think I get it now - this is the step where we convert a remote registry to a filesystem with a clone, and you are storing an old version to check if the remote should be cleanup up, and (I think?) you only had to do this because the cleanup function was removed for the local filesystem? I think I understand now but maybe we can make it more clear for future developers. Given adding the filesystem cleanup back, maybe like:

```python
# Prepare to cleanup if we clone the remote to temporary location
cleanup = False
if isinstance(remote, VersionControl):
    remote = Filesystem(remote.clone())
    cleanup = True

...
if cleanup:
    remote.cleanup()
```
It really is the same logic you have, but using a named boolean that explicitly describes what is happening (ini_remote kind of threw me through some hoops trying to figure it out!) :laughing:

muffato:
Before I go back to #590, do you think this one (and #585) could be quickly wrapped up ?

shpc/main/registry/__init__.py
muffato: Adding a cleanup boolean doesn't fully solve this, as the functions would also have to store somewhere (or return) the local registries that need to be cleaned up.

vsoch:
I don't see why not! My question (after our previous work) is if it would help to use the new Module class in here somewhere/

shpc/main/registry/remote.py
```diff
@@ -91,95 +75,106 @@ def override_exists(self, tag):
...
+    web_container_url_schemes = {
+        "github.com": "https://github.com/%s/blob/%s/%s/container.yaml",
+        "gitlab.com": "https://gitlab.com/%s/-/blob/%s/%s/container.yaml",
+    }
+
+    # The URL is substituted with repo, branch, module_name
+    raw_container_url_schemes = {
+        "github.com": "https://raw.githubusercontent.com/%s/%s/%s/container.yaml",
+        "gitlab.com": "https://gitlab.com/%s/-/raw/%s/%s/container.yaml",
+    }
```
muffato, line 95: The existence of these dictionaries is perhaps the proof I should reinstate `GitHub` and `GitLab` as sub-classes of `VersionControl` ?
And instead of exposing those strings with `%s`, they would expose methods, which would give more flexibility in case a Registry provider doesn't substitute the elements in the same order

muffato (commented):
I don't really see code here that could be simplified by using the new Module object, only perhaps a refactoring of Provider.

shpc/main/registry/remote.py
vsoch: > The existence of these dictionaries is perhaps the proof I should reinstate GitHub and GitLab as sub-classes of VersionControl ?

I like that idea I think, yes.

muffato: Done, and I like it

shpc/main/modules/base.py
```diff
@@ -235,10 +240,9 @@ def docgen(self, module_name, registry=None, out=None, branch="main"):
         aliases = config.get_aliases()
         template = self.template.load("docs.md")
         registry = registry or defaults.github_url
-        github_url = "%s/blob/%s/%s/container.yaml" % (registry, branch, module_name)
-        raw_github_url = shpc.main.registry.get_module_config_url(
-            registry, module_name, branch
-        )
+        git_registry = self.registry.get_registry(registry, tag=branch)
+        github_url = git_registry.get_container_yaml_url(module_name)
+        raw_github_url = git_registry.get_raw_container_yaml_url(module_name)
```
vsoch, line 245: Could we tweak the naming a bit, do something like:

```python
remote = self.registry.get_registry(registry, tag=branch)
github_url = remote.get_container_url(module_name)
raw_github_url = remote.get_container_url(module_name, raw=True)
```

I'm suggesting removing yaml because it's the only file we have for a container (so sort of implied?)

shpc/main/registry/__init__.py
```diff
@@ -128,19 +135,17 @@ def _sync(
...
-        remote = Remote(sync_registry, tag=tag)
-        local = self.get_registry(local or self.settings.filesystem_registry)
+        remote = self.get_registry(
+            sync_registry or self.settings.sync_registry, tag=tag
+        )
+        local = self.get_registry(local) if local else self.filesystem_registry
 
         # We sync to our first registry - if not filesystem, no go
-        if not local.is_filesystem_registry:
+        if not isinstance(local, Filesystem):
```
vsoch, line 145: This could also still be None, so maybe
```suggestion
        if not local or not isinstance(local, Filesystem):
```

shpc/main/registry/__init__.py
vsoch: Can you update me about your thinking here?

shpc/main/registry/__init__.py
```diff
@@ -171,18 +175,20 @@ def sync_from_remote(
 
         # No local registry provided, use default
         if not local:
-            local = Filesystem(self.settings.filesystem_registry)
+            local = self.filesystem_registry
```
vsoch, line 178: We will run into trouble if this is None (which it can be, will be by default!) It happens when they don't have a filesystem registry defined.

vsoch: We probably need to check for this, first thing, and cut out early if local isn't defined, or we don't have a filesystem registry (akin to sync above).

shpc/main/registry/__init__.py
vsoch: I'm of the same opinion as before - a user is likely to call "sync" and we always want to clean up the remote. A develoepr user that wants more control can use sync_from_remote, and they would then have instantiated their own local/remote and can do as they please. So I think I'd like to see:

 - sync: should call remote.cleanup()
 - sync_from_remote: should not do any cleanup (and this is what we use in testing, it's on us)

Let me know your thinking, and if it's changed from the above.

shpc/main/registry/__init__.py
```diff
@@ -171,18 +175,20 @@ def sync_from_remote(
...
 
-        tmpdir = remote.source
-        if tmpdir.startswith("http") or not os.path.exists(tmpdir):
-            tmpdir = remote.clone()
+        need_cleanup = False
```
vsoch, line 180: My preference would be to remove this cleanup logic here entirely. The developer is in control. For the sync (which calls this) that is where we clean up for them. As it is now (moving the cleanup to this level) we aren't giving them a choice anymore.

shpc/main/registry/filesystem.py
```diff
@@ -75,20 +75,31 @@ def override_exists(self, tag):
...
+        return os.path.exists(os.path.join(self.source, name))
+
     def iter_modules(self):
+        """
+        yield module names
+        """
+        # Find modules based on container.yaml
         for filename in shpc.utils.recursive_find(self.source, "container.yaml"):
-            module = os.path.dirname(filename).replace(self.source, "").strip(os.sep)
+            module = os.path.dirname(filename)
```
vsoch, line 99: This is going to miss the entire module uri (e.g., quay.io/biocontainers/samtools) and only provide "samtools"

vsoch (commented):
okay I reviewed the first part - I can come back and do the second half after we do a round on that!

shpc/main/modules/base.py
muffato: Agreed

shpc/main/registry/filesystem.py
muffato: Oups, sorry

vsoch: Haha no worries! That's what the review is for!

shpc/main/client.py
```diff
@@ -124,12 +124,16 @@ def update(self, name=None, dryrun=False, filters=None):
...
-            modules = [name]
+            # find the module in the registries. _load_container
+            # calls `container.ContainerConfig(result)` like below
+            configs = [self._load_container(name)]
         else:
-            modules = [x[1] for x in list(self.registry.iter_modules())]
-
-        for module_name in modules:
-            config = self._load_container(module_name)
+            # directly iterate over the content of the registry
```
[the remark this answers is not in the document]
muffato: Both sides are not that different. `_load_container` is just some syntactic sugar on top of `container.ContainerConfig()`. The difference (there has to be one !) is that on one side we're given a module name and we need to search it in the Registry, and on the other side the registry gives us its content directly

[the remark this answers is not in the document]
vsoch: okay, that's fair! If it's more work to provide the uri directly from the iterable again to load_container, then the different approach makes sense (or if we could get it wrong).

shpc/main/registry/filesystem.py
muffato: Can you explain the issue ? `filename` has indeed changed and is now the relative path of `container.yaml` (included) from `self.source`, but `module` is still the relative directory path that holds `container.yaml`.

muffato:
I've moved the cleanup call a level up into `_sync`

shpc/main/registry/filesystem.py
vsoch: It's hard to say without seeing how this function is changed - I think what we want here is to be able to get from:

- quay.io/biocontainers/samtools/container.yaml (fullpath or not) to
- quay.io/biocontainers/samtools

Before we were starting with the dirname of that, so:

- /full/path/modules/quay.io/biocontainers/samtools/container.yaml

and deriving

- quay.io/biocontainers/samtools

by replacing the source `/fullpath/modules` + a strip of os.sep

So perhaps a sanity check is to run this and print what you get for your new module - if it's the full `quay.io/biocontainers/samtools` then you are probably OK - it totally depends on what that (potentially now changed) function returns.

And I'm off for tonight - will make some more time tomorrow evening to review further!

muffato: I can confirm it returns "quay.io/biocontainers/samtools" (and the tests pass, and the various shpc commands I've tried work as well)

shpc/main/registry/__init__.py
```diff
@@ -171,18 +180,20 @@ def sync_from_remote(
 
         # No local registry provided, use default
```
vsoch, line 181: I think we might still miss all these checks if local is provided, but it's not a filesystem registry:
1. first check will be skipped - the path won't exist despite having a string
2. second check will be skipped - local is defined
3. we then only check if remote is a filesystem.

shpc/main/registry/remote.py
```diff
@@ -91,95 +76,81 @@ def override_exists(self, tag):
...
+            raise ValueError("Failed to clone repository {}:\n{}", self.url, e)
+        assert os.path.exists(tmpdir)
+        self._clone = Filesystem(tmpdir)
+        return self._clone
+
+    def cleanup(self):
+        """
+        Cleanup the registry
+        """
+        if self._clone:
```
vsoch, line 138: I'm someone that regularly nukes my tmpdir at random points "just for funsies" so we check for defined and still existing!

```suggestion
        if self._clone and os.path.exists(self._clone):
```

shpc/main/registry/remote.py
```diff
@@ -231,10 +205,48 @@ def iter_registry(self, filter_string=None):
...
+    def matches(cls, source):
+        return urllib.parse.urlparse(source).hostname == "github.com"
+
+    @property
+    def library_url(self):
+        url_path = self.parsed_url.path.lstrip("/")
+        if url_path.endswith(".git"):
+            url_path = url_path[:-4]
+        owner, repo = url_path.split("/", 1)
+        return f"https://{owner}.github.io/{repo}/library.json"
```
vsoch, line 226: These library_url properies are almost duplicates of one another, minus the repository name "github" or "gitlab." Should we possibly have a remote_name (or similar) attribute that goes into a shared function?

shpc/main/registry/remote.py
```diff
@@ -202,19 +176,19 @@ def _update_clone_cache(self):
         """
         logger.warning(
             "Remote %s is not deploying a Registry API, falling back to clone."
-            % self.source
+            % self.url
         )
-        tmpdir = self.clone()
-        for dirname, module in self.iter_modules():
+        tmplocal = self.clone()
```
vsoch, line 181: It looks like we are switching between tmplocal and self commands - e.g., the self will hit `self._cache` and tmplocal is a direct pointer to it. This can be easy for the developer to forget (e.g., why did we do self.cleanup when we were working with tmplocal? I don't have preference for which handle is used, but could we be consistent to use the same one in the various spots of this function? :thinking:

shpc/main/registry/remote.py
```diff
@@ -231,10 +205,48 @@ def iter_registry(self, filter_string=None):
...
+
+    @property
+    def library_url(self):
+        url_path = self.parsed_url.path.lstrip("/")
+        if url_path.endswith(".git"):
+            url_path = url_path[:-4]
+        owner, repo = url_path.split("/", 1)
+        return f"https://{owner}.gitlab.io/{repo}/library.json"
+
+    def get_container_url(self, module_name):
```
vsoch, line 248: These are different enough that I like how they are separate - thanks for doing this change I also like it better!

vsoch (commented):
ok we're close!

shpc/main/registry/__init__.py
muffato: Some of those checks were in `_sync`, but they're indeed more useful here. I'll move them

shpc/main/registry/remote.py
muffato: I didn't do exactly that, but I could shorten both implementations of `library_url` quite a bit.
I want to avoid the base-class knowing too much about its sub-classes and how GitHub/GitLab work

vsoch:
@muffato this was my contribution (and mistake) - I think self._cache is a FIlesystem so instead of:

```
  if self._clone and os.path.exists(self._clone)
```
We need to either have an exists function on that (e.g., self._clone.exists()) or reference the path attribute for it directly.

muffato:
My bad as well. I should have noticed before committing the change

muffato:
Yeah, the tests pass 😎 !

shpc/main/registry/__init__.py
```diff
@@ -163,26 +162,39 @@ def sync_from_remote(
...
+            local = self.filesystem_registry
+            # We sync to our first registry - if not filesystem, no go
+            if not local:
+                logger.exit("No local registry to sync to. Check the shpc settings.")
 
-        tmpdir = remote.source
-        if tmpdir.startswith("http") or not os.path.exists(tmpdir):
-            tmpdir = remote.clone()
+        if not isinstance(local, Filesystem):
+            logger.exit("Can only synchronise to a local file system, not to %s." % local)
```
vsoch, line 181: Sorry I know US English is terrible - we use z!
```suggestion
            logger.exit("Can only synchronize to a local file system, not to %s." % local)
```

vsoch (commented):
One quick question - where does the ssh come in (have you tested and it works for you?)

After that, just a note in a the CHANGELOG and version bump and we are good to go!

muffato:
> One quick question - where does the ssh come in (have you tested and it works for you?)

It still works. I've tried `shpc sync-registry`, `shpc show`, `shpc install`, and `shpc docgen`

The Registry has to be given as `ssh://git@github.com/singularityhub/shpc-registry.git` which is *not* one what you get by default from the GitHub UI. It needs the extra leading `ssh://` and a `/` instead of `:` after the hostname.

There isn't any obvious code to support ssh because 1) it's still a URL, though on a different scheme, so `urllib` works as usual, and 2) the URL is passed on to `git clone` which supports all.
Note that even with ssh, shpc still tries to access a library.json, which in this case works. But when the library.json is not accessible, then it falls back to the clone and a local traversal.

I need to add some tests to make it obvious ssh is supported.

muffato:
I've added a SSH url to the tests.

The changelog and version bump will have to happen in the parent PR: #585 , where the merge conflict with `main` has to be handled.

vsoch:
oh boop, I totally forgot this was a nested PR! Thanks for your patience with me @muffato !

vsoch (approved):

PR #590, opened 2022-09-19
Reinstallation mode

### LINKED ISSUE #501: Command to rebuild all modules / wrapper scripts

muffato:
Closes #501 

Hi @vsoch . Working my way through #501. It's not as straightforward as I though it would be !

This is a draft pull-request. So far I've expanded `shpc install` to do clean re-installations, and I've got a stub function to identify what "reinstall all" will do. Can I get feedback on this first ?


shpc/main/modules/base.py
```diff
@@ -480,3 +497,49 @@ def install(
...
+        """
+        modules = self._get_module_lookup(
+            self.settings.module_base, self.modulefile, pattern
+        )
+
+        # If we don't have modules, exit early
+        if not modules:
+            logger.exit("You don't have any install modules. Try shpc show.", 0)
+
+        unavailable_modules = False
```
vsoch, line 515: Ok this is getting a bit hairy - my instinct is saying we want to break this down into a few steps.

1. get a list of installed modules and all modules. We can derive those already installed vs. not installed yet with basic sets, and OK to add a version to that if desired.
2. I think we probably want to just call the self.install function directly from self.reinstall
3. I think we should assume a reinstall all is going for the latest version - no need to check and spit out a warning. The one warning we want (and a continue in the loop) is if something is installed but we don't have a source anymore. In this case it's just ignored (not deleted)
4. And there might be a `def reinstall` function to handle operations for a single case, with optional arguments for the lookup of already installed vs. known, and then someone could call it one off for one package:

```bash
$ shpc reinstall python
This will reinstall python@3.5.1 as python@3.9.1 are you sure [Y/N]?
```

And then the reinstall all case:

```bash
$ shpc reinstall --all
```
And they would be using the same underlying functions. And even for the large "all" case we might want to do this interactively, where we essentially create a table of installed vs will be installed (with empty spaces where appropriate) and then ask the user (as we do above) y/n "Do you want to continue?" If no, they might have the choice to specify a specific version pin, e.g.,

```bash
$ shpc reinstall --all --pin python@3.5.1
$ shpc reinstall --all -p python@3.5.1
```
If you are overwhelmed (or just short of time) I can probably take a crack this weekend! I think this feature will be great to have! Let me know if you want more detail or pseudocode for any of the above!

shpc/tests/test_client.py
```diff
@@ -338,3 +336,41 @@ def test_add(tmp_path, module_sys, remote):
         client.get("dinosaur/salad:latest")
     client.install("dinosaur/salad:latest")
     assert client.get("dinosaur/salad:latest")
+
+
+@pytest.mark.parametrize(
```
vsoch, line 341: You are so good at writing tests! I'm bad I always do this last. I admire that.

vsoch (commented):
Here is some quick feedback!

shpc/main/modules/base.py
muffato: Thanks for the feedback. Maybe the name `reinstall` isn't quite right. The use case I've got in mind is to regenerate all modules & co when shpc itself has a new feature, e.g. wrapper scripts for aliases, wrapper scripts for helpers, etc, while leaving all the versions as they are. I would disagree with "we should assume a reinstall all is going for the latest version". But I agree a general upgrade (like `brew upgrade`) would be nice to add.
Would you be happy making the argument of `--pin` optional ? therefore allowing `shpc reinstall --all --pin` to fulfil the scenario I've got in mind. I actually already had a commit that adds some CLI parameters for reinstall, and I didn't push it because I wanted to think more about the CLI interface, e.g. the `shpc reinstall python` you're proposing would do the same as `shpc install python` and that'd confuse me as a user.

I'm unlikely to have any time this week to implement this, but I can still discuss the feature

shpc/tests/test_client.py
muffato: Honestly, I'm doing it here because I saw you doing it. Peer pressure ;) !

shpc/main/modules/base.py
vsoch: Ah I think I better understand now!

> Would you be happy making the argument of --pin optional ?

Yes absolutely - and if what I'm suggesting would just emulate `shpc upgrade --all` then perhaps we don't need `--pin` (and the behavior you describe would be the default.

> shpc reinstall python you're proposing would do the same as shpc install python and that'd confuse me as a user.

The difference is that `shpc install` would get latest, and `shpc reinstall` would try to re-create the existing version.

muffato:
Hi. FYI I've got `shpc reinstall` and `shpc upgrade` working. The code is very fresh and rough and I want to work more on it, but the core is there.

vsoch:
Nice! Let me know when you’d like a review or want help debugging tests.

muffato:
I feel much more comfortable than when I started, and I think I understand the corner cases.

Could you please review the following ?

- `shpc install` takes a tool name, and optionally a version. If no version is given, defaults to the latest one.
  - if the tool/version is not yet installed, install as usual
  - if the tool/version is already installed: quit, unless `--force` is given. In which case, do a proper uninstallation first and then the regular install
  - I will change the `view` parameter so that multiple views can be requested, which will allow the reinstallation to restore the tool in all necessary views.
- `shpc reinstall` takes either a tool name (possibly versioned) or `--all`
  - scope:
    - if one tool and one version is requested: only do this one version
    - if one tool is requested, but no specific version: do all its versions that are installed
    - if `--all` is given, list all the tools and versions that are installed
  - procedure:
    1. check whether we still have the required recipes in the registry.
    2. bail out if a recipe is missing, unless `--force`
    3. call `install` on all required recipes in "force" mode to force the reinstallation, in the same views as they were
- `shpc upgrade` takes either a tool name (cannot be versioned) or `--all`
  - scope:
    - if one tool is requested: only do this one
    - if `--all` is given, list all the tools that are installed
  - procedure:
    1. check whether we still have recipes for those tools in the registry.
    2. bail out if a recipe is missing, unless `--force`
    3. identify the latest version of each recipe
    4. install the latest versions, in "force" mode to force a reinstallation if necessary. The latest version is added to all the views that had _any_ version of that tool
    5. uninstall all the other installed versions

vsoch:
> shpc install takes a tool name, and optionally a version. If no version is given, defaults to the latest one.

correct, where latest is determined by the container.yaml.

> if the tool/version is not yet installed, install as usual

yes!

> if the tool/version is already installed: quit, unless --force is given. In which case, do a proper uninstallation first and then the regular install

yes, although there might be a nice human-friendly error message that says "X is already installed - use --force to uninstall and reinstall." Basically, the user always knows exactly what is going to happen (e.g., doesn't assume it will reinstall the same version, etc.)

> I will change the view parameter so that multiple views can be requested, which will allow the reinstallation to restore the tool in all necessary views.

Can you show me what this would look like on the command line? E.g., right now we have:

```bash
   shpc view install <view> <module>
$ shpc view install mpi ghcr.io/autamus/clingo
```
And I'm not seeing how this could be made to accept multiple views without getting ugly/hairy.

> shpc reinstall takes either a tool name (possibly versioned) or --all

Okay so:

```bash
$ shpc reinstall --all
```
Case 1: missing a tool in the registry:
```console
We cannot find <module> to reinstall. Use force to continue and permanently uninstall <module>.
```
And then it seems like there is no option to force but leave the module?

For the rest of reinstall and upgrade I'm having trouble understanding your notes (e.g., with scope and procedure) could you write out the command options and then the possible outcomes verbatim? E.g., it's not clear to me why reinstall/upgrade would do a listing, but if you write out the interaction in more detail I think I'll understand.

muffato:
> Can you show me what this would look like on the command line? E.g., right now we have:
> 
> ```shell
>    shpc view install <view> <module>
> $ shpc view install mpi ghcr.io/autamus/clingo
> ```
> 
> And I'm not seeing how this could be made to accept multiple views without getting ugly/hairy.

I was only thinking of changing `shpc install` to allow this:
```shell
shpc install quay.io/biocontainters/samtools --view view1 --view view2
```
Right now, one would have to run two commands:
```shell
shpc install quay.io/biocontainters/samtools --view view1
shpc view install view2 quay.io/biocontainters/samtools
```
-----

> Okay so:
> 
> ```shell
> $ shpc reinstall --all
> ```
> 
> Case 1: missing a tool in the registry:
> 
> ```
> We cannot find <module> to reinstall. Use force to continue and permanently uninstall <module>.
> ```
> 
> And then it seems like there is no option to force but leave the module?

That's indeed the question: what should happen if a tool and/or version is not found in the registry ? Should it be uninstalled or left there ?
Maybe, rather than `--force`, the CLI options could explicitly let the user decide what to do: no option (= abort) vs `--ignore-missing` vs `--uninstall-missing`

---

> For the rest of reinstall and upgrade I'm having trouble understanding your notes (e.g., with scope and procedure) could you write out the command options and then the possible outcomes verbatim? E.g., it's not clear to me why reinstall/upgrade would do a listing, but if you write out the interaction in more detail I think I'll understand.

What I described is how the command should work internally, as essentially wrappers around other existing internal methods  of shpc, namely: list installed modules, find modules in the registry, install a module, and uninstall a module. The commands are not meant to be showing their inner working to the user.
As wrappers, the only thing they may print themselves on stdout/stderr would be info/warning about missing modules/versions. But a lot more may be written on stdout/stderr by the existing `install` and `uninstall` methods if I use them as-is, e.g.:
```
singularity pull --name /nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/containers/quay.io/biocontainers/samtools/1.12--h9aed4be_1/quay.io-biocontainers-samtools-1.12--h9aed4be_1-sha256:5fd5f0937adf8a24b5bf7655110e501df78ae51588547c8617f17c3291a723e1.sif docker://quay.io/biocontainers/samtools@sha256:5fd5f0937adf8a24b5bf7655110e501df78ae51588547c8617f17c3291a723e1
INFO:    Using cached SIF image
Module quay.io/biocontainers/samtools:1.12--h9aed4be_1 was created.
Creating link $module_base/quay.io/biocontainers/samtools/1.12--h9aed4be_1/module.lua -> $views_base/view1/samtools/1.12--h9aed4be_1.lua
```
or
```
$container_base/quay.io/biocontainers/samtools/1.12--h9aed4be_1 and all subdirectories have been removed.
$module_base/quay.io/biocontainers/samtools/1.12--h9aed4be_1 and all subdirectories have been removed.
/nfs/users/nfs_m/mm49/workspace/tol-it/central_software/singularity-hpc/views/view1/samtools/1.12--h9aed4be_1.lua has been removed.
```

Imagine that for dozens of modules ! I was considering adding a `quiet` flag in these to skip those messages and instead have just one line per module, e.g.
```
$ shpc reinstall quay.io/biocontainers/samtools:1.12--h9aed4be_1
Module quay.io/biocontainers/samtools:1.12--h9aed4be_1 was reinstalled.
$ shpc reinstall quay.io/biocontainers/samtools
Module quay.io/biocontainers/samtools:1.12--h9aed4be_1 was reinstalled.
Module quay.io/biocontainers/samtools1.13--h8c37831_0 was reinstalled.
$ shpc reinstall --all
Module quay.io/biocontainers/samtools:1.12--h9aed4be_1 was reinstalled.
Module quay.io/biocontainers/samtools1.13--h8c37831_0 was reinstalled.
Module quay.io/biocontainers/bcftools:1.14--h88f3f91_0 was reinstalled.
```
or
```
$ shpc upgrade quay.io/biocontainers/samtools
Module quay.io/biocontainers/samtools was upgraded to 1.14--hb421002_0
$ shpc reinstall --all
Module quay.io/biocontainers/samtools was upgraded to 1.14--hb421002_0
Module quay.io/biocontainers/bcftools was upgraded to 1.14--h88f3f91_0
```

vsoch:
> I was only thinking of changing shpc install to allow this:
> shpc install quay.io/biocontainters/samtools --view view1 --view view2

Hmm - so I don't like that we would provide two ways to do the same thing - that's a confusing client interaction. I also like the consistency of `shpc view <command>`. If someone really wants to do multiple views, they can just do:

```bash
for view in mpi1 mpi2
  do
    shpc view install $view quay.io/biocontainters/samtools
done
```
But to step back a second - views are symlinks. There isn't really a concept of reinstall that doesn't alter a view. We really should just be updating the views with respect to their reinstall requests, the reason being that there is no concept of doing a reinstall (with a sylinked view) without completely changing that, e.g., "keep the view but do the reinstall" is not possible, so adding/not adding the --view flag is irrelevant. Here is an example interaction that might make sense:

I am starting with samtools installed, and it happens to be under two views (two versions)

```bash
$ shpc reinstall quay.io/biocontainters/samtools
```
```console
You currently have:
   samtools@v1 (view1)
   samtools@v2
   samtools@v3 (view2)
Are you sure you would like to reinstall all versions to update all views (y/n)?
```
And then they would be like "OH" I really just want to do that specific version...

```bash
$ shpc reinstall quay.io/biocontainters/samtools@v1
```
```console
You currently have:
   samtools@v1 (view1)
Are you sure you would like to reinstall all versions to update all views (y/n)?
```
For the case where someone wants to change a version installed to a view, this would fall under an upgrade (and not reinstall, if I understand the distinction).

```bash
#  Case 1: missing a tool in the registry:
$ shpc reinstall --all
We cannot find <module> to reinstall. Use force to continue and permanently uninstall <module>.
```
Perfecto. That is clear - if you want to reinstall all and you are missing something? It just gets removed.

> That's indeed the question: what should happen if a tool and/or version is not found in the registry ? Should it be uninstalled or left there ?

I do think this would be a reasonable desire (not to remove) and I like your suggestion, e.g.,:

```bash
#  Case 1: missing a tool in the registry:
$ shpc reinstall --all
We cannot find <module> to reinstall! Your choices:
> Use --uninstall-missing to permanently uninstall <module> (this would be akin to force, but more specific)
> Use --ignore-missing to uninstall only those that are found and ignore (keep) those that are missing.
```
And please tweak the language above as you see fit! I think as long as we give a clear message of the current state, and the options available to take, it's a fairly good user interaction.

And yes ! I definitely like the quiet command. It's more verbose now to show exactly what is happening, although that was intended for just one install (and not this multi case) so a `--quiet` flag would make sense.

muffato:
> But to step back a second - views are symlinks. There isn't really a concept of reinstall that doesn't alter a view. We really should just be updating the views with respect to their reinstall requests, the reason being that there is no concept of doing a reinstall (with a sylinked view) without completely changing that, e.g., "keep the view but do the reinstall" is not possible, so adding/not adding the --view flag is irrelevant. Here is an example interaction that might make sense:

And we agree on that. My `reinstall` and `upgrade` commands don't have `--view` options. The example I had there was `shpc install` which _already_ has `--view`.
As you said, `reinstall` needs to preserve the views. For that, the most practical is to make the internal `install()` method understand multiple view requests (currently it only allows one). And once that's implemented, I would as well expose it in `shpc install` itself.
Re. your comment about the client having multiple ways of doing things: then there's already a problem ! Right now, one can do
```
shpc install $module
shpc view install $view $module
```
or
```
shpc install --view $view $module
```
Would you want to remove the `--view` option of `shpc install` ?

vsoch:
Gotcha - I missed that it was just for install.

```
shpc install $module
shpc view install $view $module

or

shpc install --view $view $module
```
These are subtly different - the first is installing to core shpc, the second is just interacting with a view, and the third is installing (and adding to a view to save time). As a user, at least, I'd expect to have both of those. If anything, I'd probably update the client to remove the `--view` flag entirely to be more consistent.

muffato:
Hi @vsoch . I found that `shpc view install $view $module` too does a core installation transparently, and it is a design choice: https://github.com/singularityhub/singularity-hpc/blob/main/shpc/client/view.py#L167-L168
which gives 3 ways of installing a module into a view
```
# Intended method
shpc install $module
shpc view install $view $module
# Shortcut n˚1
shpc view install $view $module
# Shortcut n˚2
shpc install --view $view $module
```

But, more importantly, both shortcuts (`shpc view install $view $module` and `shpc install --view $view $module`) _ignore_ the `default_view` setting when doing the core installation.

My view on this:

- I would first make sure the `default_view` setting is always honoured.
- I'm OK with `view install` doing a core installation, but can remove that behaviour if you want.
- I dislike the `--view` option of `shpc install` accepting a _single_ value. Either the option is removed altogether, or it can be repeated so that one can list _all_ the views to add the module to.

vsoch:
Yes let’s remove the third option I think!

muffato:
Still some work to do (and I need to fix the tests !) but `shpc reinstall` works. Could I get a review on it, @vsoch ?

There's also one difficulty I haven't managed to overcome. Modules that were installed against a preexisting container image won't match their `container.yaml` and ca't be reinstalled the normal way. How would you suggest dealing with that ? Parsing the existing module to find out the path to the container ?

shpc/client/install.py
```diff
@@ -36,5 +35,4 @@ def main(args, parser, extra, subparser):
             cli.settings.default_view,
             args.install_recipe,
             force=args.force,
-            container_image=args.container_image,
```
vsoch, line 39: okay, and this will work because we are symlinking the module file, which has the path to the container, wherever it may be!

shpc/client/reinstall.py
```diff
@@ -0,0 +1,46 @@
+__author__ = "Vanessa Sochat"
+__copyright__ = "Copyright 2021-2022, Vanessa Sochat"
```
vsoch, line 2:
```suggestion
__copyright__ = "Copyright 2022, Vanessa Sochat"
```

shpc/client/reinstall.py
```diff
@@ -0,0 +1,46 @@
...
+        module_sys=args.module_sys,
+        container_tech=args.container_tech,
+    )
+
+    # Update config settings on the fly
+    cli.settings.update_params(args.config_params)
+
+    # It doesn't make sense to give a module name and --all
+    if args.reinstall_recipe and args.all:
+        logger.exit("Conflicting arguments reinstall_recipe and --all, choose one.")
```
vsoch, line 27: I don't think the user knows what the argument name is, at least without looking.

```suggestion
        logger.exit("Conflicting arguments: choose a single module name to reinstall OR --all.")
```

shpc/client/reinstall.py
```diff
@@ -0,0 +1,46 @@
...
+    # One option must be present
+    if not args.reinstall_recipe and not args.all:
+        logger.exit("Missing arguments: provide reinstall_recipe or --all.")
```
vsoch, line 30:
```suggestion
        logger.exit("Missing arguments: provide a single module to reinstall OR --all.")
```

shpc/client/reinstall.py
```diff
@@ -0,0 +1,46 @@
...
+    if args.ignore_missing and args.uninstall_missing:
+        logger.exit(
+            "Conflicting arguments --ignore-missing and --uninstall-missing, choose one."
```
vsoch, line 33:
```suggestion
            "Conflicting arguments: choose one of --ignore-missing OR --uninstall-missing."
```

shpc/main/modules/base.py
```diff
@@ -375,15 +375,19 @@ def get_module(self, name, container_image=None, keep_path=False):
         return module
 
     def install(
-        self, name, force=False, container_image=None, keep_path=False, **kwargs
+        self,
+        name,
+        allow_reinstall=False,
```
vsoch, line 380: As I mentioned above, let's make this `force`
```suggestion
        force=False,
```

shpc/main/modules/base.py
```diff
@@ -375,15 +375,19 @@ def get_module(self, name, container_image=None, keep_path=False):
...
+        container_image=None,
+        keep_path=False,
+        **kwargs
     ):
         """
         Given a unique resource identifier, install a recipe.
 
         For lmod, this means creating a subfolder in modules, pulling the
         container to it, and writing a module file there. We've already
         grabbed the name from docker (which is currently the only supported).
```
vsoch, line 390:
```suggestion
        grabbed the name from docker (which is currently the only supported).
        A "force" indicates a reinstall will happen.
```

shpc/main/modules/base.py
```diff
@@ -393,6 +397,23 @@ def install(
         # We always load overrides for an install
         module.load_override_file()
 
+        # Check previous installations of this module
+        if os.path.exists(module.module_dir):
+            if not allow_reinstall:
```
vsoch, line 402:
```suggestion
            if not force:
```

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
         # Don't continue if it exists, unless force is True
         view.confirm_install(module.module_dir, force=force)
         view.install(module.module_dir)
+
+    def reinstall(self, name, when_missing=None):
+        """
+        Reinstall the module, or all modules
+        """
+        if name:
+            module_name, _, version = name.partition(":")
```
vsoch, line 470: Is there some redundancy here with functionality - could we parse the name using our Module class?

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
...
+                version and version not in installed_modules[module_name]
+            ):
+                logger.exit("%s is not installed. Nothing to reinstall." % name)
+            versions = [version] if version else installed_modules[module_name]
+            # Reinstall the required version(s) one by one
+            for version in versions:
+                self._reinstall(module_name, version, when_missing)
+        else:
+            # Reinstall everything that is currently installed
+            installed_modules = self._get_module_lookup(
```
vsoch, line 485: Hmm I wonder if there is some way to simplify the logic here (and not call the function in both blocks but use a shared call?) See duplicate of (almost) identical code always makes me pause!

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
...
+            )
+            if (module_name not in installed_modules) or (
+                version and version not in installed_modules[module_name]
+            ):
+                logger.exit("%s is not installed. Nothing to reinstall." % name)
+            versions = [version] if version else installed_modules[module_name]
+            # Reinstall the required version(s) one by one
+            for version in versions:
+                self._reinstall(module_name, version, when_missing)
+        else:
```
vsoch, line 483: okay I see what you are doing - without a name we hit the else and this is "reinstall all." I will think more about this one - I don't have a better design idea at 5:30am! But not reading the code, it's not obvious this function handles single / all use cases (beyond the comment in the docstring). I think given that we are ultimately looping through a list of modules (and the versions come from a lookup) or would otherwise just be None) it might be nice to have logic to populate the list of modules and versions, and one common loop through them to reinstall. The other option is to have two clearly distinguished functions, but that might be overkill because we already have two!

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
...
+        """
+        result = self.registry.find(module_name)
+        if result:
+            config = container.ContainerConfig(result)
+            if version in config.tags:
+                return self.install(module_name + ":" + version, allow_reinstall=True)
+            else:
+                missing = module_name + ":" + version
+        else:
+            missing = module_name
```
vsoch, line 504: The double nested if - is there a better way to write this? I think we can just plop the entire thing (name and version) into missing? And since we return above we don't need the elses.

```suggestion
missing = (module_name + ":" + version) if version else module_name
```

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
...
+
+    def _reinstall(self, module_name, version, when_missing):
+        """
+        Reinstall (and possibly upgrade) all the current modules, possibly filtered by pattern.
+        """
+        result = self.registry.find(module_name)
+        if result:
+            config = container.ContainerConfig(result)
+            if version in config.tags:
+                return self.install(module_name + ":" + version, allow_reinstall=True)
```
vsoch, line 500:
```suggestion
                return self.install(module_name + ":" + version, force=True)
```

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
...
+            else:
+                missing = module_name + ":" + version
+        else:
+            missing = module_name
+
+        if when_missing:
+            if when_missing == "ignore":
+                logger.info(
+                    "%s is not in the Registry any more. Ignoring as instructed."
```
vsoch, line 509:
```suggestion
                    "%s is not in the registry any more. Ignoring as instructed."
```

shpc/main/modules/base.py
```diff
@@ -440,3 +461,58 @@ def view_install(self, view_name, name, force=False, container_image=None):
...
+        if result:
+            config = container.ContainerConfig(result)
+            if version in config.tags:
+                return self.install(module_name + ":" + version, allow_reinstall=True)
+            else:
+                missing = module_name + ":" + version
+        else:
+            missing = module_name
+
+        if when_missing:
```
vsoch, line 506: Can we collapse these into a single if elif elif, e.g.,:

```python
if not when_missing:
     ...
elif when_missing == "ignore":
    ...
elif when_missing == "uninstall":

else:
   # maybe mention whatever they provided isn't a valid option?
```

vsoch (commented):
okay review in! I'll ignore failed tests for now since we are still discussing design. Thanks @muffato !

vsoch:
> Modules that were installed against a preexisting container image won't match their container.yaml and ca't be reinstalled the normal way. How would you suggest dealing with that ? Parsing the existing module to find out the path to the container ?

We could certainly try! If we can clean up the module files a bit to make it easy to parse (and find the container path), this could actually be feasible do do. Would we honor the previous location? Pull a new one? I think there are two cases:

- Reinstall updated version - and for this I'm tempted to say if it's not in the module /container folder maintained by shpc, to just leave it there. E.g., imagine the BioContainers case - we wouldn't want to try and delete / repull down to there. We'd probably try a repull, and if that fails, we could either give an error message or allow the user to use the previous SIF
- Reinstall the same version - parse the module file to get the exact container path (and re-use). This would maybe be the fallback for the case above.

Let me know what you think!

muffato:
Closing as superseded by #674

PR #594, opened 2022-10-09
Run pyflakes on everything except __init__.py and shell.py

### LINKED ISSUE #592: pyflakes
Hi @vsoch 

As I was running tests, I realised that pyflakes is run on certain files only: https://github.com/singularityhub/singularity-hpc/blob/main/.github/workflows/main.yml#L33-L62 . Most files actually pass pyflakes, and I can make a PR to update the CI.

Just one question about this one: https://github.com/singularityhub/singularity-hpc/blob/main/shpc/client/shell.py#L52-L59
pyflakes complains that the `client` variable is not used anywhere: should it be used in the `embed()` call like for bpython below ?
Otherwise, I'll call `create_client` without storing the return value in `client` to make pyflakes happy

muffato:
Closes #592

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,29 +34,4 @@ jobs:
         run: |
           export PATH="/usr/share/miniconda/bin:$PATH"
           source activate black
-          pyflakes shpc/utils/fileio.py
-          pyflakes shpc/utils/terminal.py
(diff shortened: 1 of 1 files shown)


vsoch (approved):
This is so much nicer! Thank you @muffato !

PR #596, opened 2022-10-10
View install commands

muffato:
Hi.

Here is a pull-request to address https://github.com/singularityhub/singularity-hpc/pull/590#issuecomment-1272536452 and https://github.com/singularityhub/singularity-hpc/pull/590#issuecomment-1272536735

1. `shpc install` doesn't accept a `--view` argument anymore.
2. `shpc view install` now honours the default view, as set in the settings.
3. I've renamed the `view` and `disable_view` arguments of the `install` method to resp. `extra_view` and `disable_default_view` to clarify what they refer to.
4. I've updated the CI to test views on tcsh too, so that both shells test (almost[^1]) entirely the same thing.
5. Both CI tests clean up their modules / views to ensure nothing bleeds from the first one to the next

[^1]: tcsh still doesn't run `shpc test`. I've tried to add it, but couldn't manage to make it work: it complains that `module` is not found even though the module commands work. I found something weird: `shpc test` [always runs the test script with bash](https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/modules/base.py#L162-L169) even if the user requested a different test shell. I tried using the test shell throughout but it didn't help. And anyway, tcsh is [not allowed](https://github.com/singularityhub/singularity-hpc/blob/e04f40646879c8b4b9810cc598f45c1319386f67/shpc/main/schemas.py#L134) as a shell.

### DIFF
--- a/shpc/main/modules/base.py
+++ b/shpc/main/modules/base.py
@@ -5,7 +5,6 @@ __license__ = "MPL 2.0"
 import inspect
 import json
 import os
-import shutil
 import subprocess
 import sys
 from datetime import datetime
@@ -19,6 +18,7 @@ import shpc.main.registry as registry
 import shpc.utils as utils
 from shpc.logger import logger
 from shpc.main.client import Client as BaseClient
+from .module import Module
 
 
 class ModuleBase(BaseClient):
@@ -72,13 +72,7 @@ class ModuleBase(BaseClient):
         Given a unique resource identifier, uninstall a module. If a view
         name is provided, assume we only want to uninstall from the view
         """
-        # The name can either be a folder or an install directory
-        name = self.add_namespace(name)
-
-        # folder paths do not have :
-        name = name.replace(":", os.sep)
-        module_dir = os.path.join(self.settings.module_base, name)
-        container_dir = self.container.container_dir(name)
+        module = self.new_module(name)
 
         # We need to look for the module in all views and show to the user first
         views_with_module = set()
@@ -88,7 +82,7 @@ class ModuleBase(BaseClient):
 
             # If uninstalling the entire module, clean up symbolic links in all views
             for view_name, entry in self.views.items():
-                if entry.exists(module_dir):
+                if entry.exists(module.module_dir):
                     views_with_module.add(view_name)
 
         # Ask before deleting anything!
@@ -106,29 +100,35 @@ class ModuleBase(BaseClient):
         if view:
             if view not in self.views:
                 logger.exit("View %s does not exist, cannot uninstall." % view)
-            return self.views[view].uninstall(module_dir)
+            return self.views[view].uninstall(module.module_dir)
 
         # Podman needs image deletion
-        self.container.delete(name)
+        self.container.delete(module.name)
 
-        if container_dir != module_dir:
+        if module.container_dir != module.module_dir:
             self._uninstall(
-                container_dir, self.container_base, "$container_base/%s" % name
+                module.container_dir,
+                self.container_base,
+                "$container_base/%s" % module.name,
             )
             self._uninstall(
-                module_dir, self.settings.module_base, "$module_base/%s" % name
+                module.module_dir,
+                self.settings.module_base,
+                "$module_base/%s" % module.name,
             )
         else:
             self._uninstall(
-                module_dir, self.settings.module_base, "$module_base/%s" % name
+                module.module_dir,
+                self.settings.module_base,
+                "$module_base/%s" % module.name,
             )
 
         # If uninstalling the entire module, clean up symbolic links in all views
         for view_name, view in self.views.items():
-            view.uninstall(module_dir)
+            view.uninstall(module.module_dir)
 
         # parent of versioned directory has module .version
-        module_dir = os.path.dirname(module_dir)
+        module_dir = os.path.dirname(module.module_dir)
 
         # update the default version file, if other versions still present
         if os.path.exists(module_dir):
@@ -331,21 +331,35 @@ class ModuleBase(BaseClient):
         at updates for entire tags. If a specific folder is provided with
         a container, check the digest.
         """
-        module_name = self.add_namespace(module_name)
-
-        # If a tag is provided, convert to directory
-        module_dir = module_name.replace(":", os.sep)
-
-        # We derive the current version installed from the container
-        # We assume the user has provided the correct prefix
-        module_dir = os.path.join(self.settings.module_base, module_dir)
-        if not os.path.exists(module_dir):
+        module = self.new_module(module_name)
+        if not os.path.exists(module.module_dir):
             logger.exit(
-                "%s does not exist. Is this a known registry entry?" % module_dir
+                "%s does not exist. Is this a known registry entry?" % module.module_dir
             )
 
-        config = self._load_container(module_name.rsplit(":", 1)[0])
-        return self.container.check(module_name, config)
+        return module.check()
+
+    def new_module(self, name, tag=None, tag_exists=True):
+        """
+        Create a new module
+        """
+        name = self.add_namespace(name)
+
+        # If the module has a version, overrides provided tag
+        if ":" in name:
+            name, tag = name.split(":", 1)
+
+        module = Module(name)
+        module.config = self._load_container(module.name, tag)
+
+        # Ensure the tag exists, if required, uses config.tag
+        if tag_exists:
+            module.validate_tag_exists()
+
+        # Pass on container and settings
+        module.container = self.container
+        module.settings = self.settings
+        return module
 
     def install(self, name, tag=None, force=False, **kwargs):
         """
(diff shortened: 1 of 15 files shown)


shpc/main/modules/base.py
```diff
@@ -348,7 +348,13 @@ def check(self, module_name):
         return self.container.check(module_name, config)
 
     def install(
-        self, name, tag=None, view=None, disable_view=False, force=False, **kwargs
+        self,
+        name,
+        tag=None,
+        extra_view=None,
```
vsoch, line 354: If we really wanted to make this an extra thing, we could remove mentions of view from the named arguments and add to kwargs instead, e.g., `extra_view = kwargs.get("extra_view")`. And I'd be OK with still just using `view` for the argument name, but having it populate into `extra_view` so it's more clear for the developer, e.g.,:

```python
extra_view = kwargs.get("view")
```
I kind of like that, because the "default view" comes from settings (unknown possibly to the function caller, who might just want to define a `view`) but for the developer reader, it is nicer to read the variable as `extra_view`. Two birds, one stone!

shpc/main/modules/base.py
```diff
@@ -387,22 +393,27 @@ def install(
         container_dir = self.container.container_dir(subfolder)
 
         # Are we also installing to a named view?
-        if view is None and not disable_view:
-            view = self.settings.default_view
+        view_names = []
```
vsoch, line 396: Hmm.. so I think allowing two views here is confusing. It should be either or (how it was before). because If we allow multiple the next logical question is "why can't we provide more than one as an argument?" and I don't want to support that!

shpc/main/modules/base.py
```diff
@@ -387,22 +393,27 @@ def install(
...
+        if extra_view:
+            view_names.append(extra_view)
 
         # We only want to load over-rides for a tag at install time
         config.load_override_file(tag.name)
 
         # A view is a symlink under views_base/$view/$module
-        if view:
-            if view not in self.views:
+        for view_name in view_names:
```
vsoch, line 406: And then this gets restored to the original logic.

shpc/main/modules/base.py
```diff
@@ -387,22 +393,27 @@ def install(
...
                 logger.exit(
-                    "View %s does not exist, shpc view create %s." % (view, view)
+                    "View %s does not exist, shpc view create %s."
+                    % (view_name, view_name)
                 )
 
-            # Update view from name to be View to interact with
-            view = self.views[view]
+        # Update view from name to be View to interact with
+        views = [self.views[view_name] for view_name in view_names]
```
vsoch, line 414: And here too.

shpc/main/modules/base.py
```diff
@@ -476,7 +487,7 @@ def install(
         logger.info("Module %s was created." % name)
 
         # Install the module (symlink) to the view and create version file
-        if view:
+        for view in views:
```
vsoch, line 490: And here too.

shpc/tests/test_views.py
```diff
@@ -76,7 +76,7 @@ def test_views(tmp_path, module_sys, module_file, container_tech, remote):
     assert not view._config["view"]["system_modules"]
 
     # Now install to it via the client
-    client.install("ghcr.io/autamus/emacs:27.2", view=view_name)
+    client.install("ghcr.io/autamus/emacs:27.2", extra_view=view_name)
```
vsoch, line 79: Perhaps just `view` here that is a kwarg, and that populates into a variable `extra_view`?

vsoch (commented):
Thanks @muffato ! Some comments below!

shpc/main/modules/base.py
muffato: It's still allowed here because this function is what `shpc view install` calls. In other words, this function here needs to  do i) core installation, ii) default view installation, iii) custom view installation.

My opinion is that this method shouldn't be dealing with views at all: just doing core installation. The reason I didn't move view installation away is that it relies on `View` objects that are only instantiated in this class. Moving everything will make the diff larger.

Note that for the reinstall feature, I will need support _somewhere_ for multiple views.

muffato:
The latest version makes most of the comments obsolete

shpc/main/modules/base.py
```diff
@@ -475,8 +456,47 @@ def install(
             name = "%s:%s" % (name, tag.name)
         logger.info("Module %s was created." % name)
 
-        # Install the module (symlink) to the view and create version file
-        if view:
-            view.install(module_dir)
-
         return container_path
+
+    def view_install(self, view_name, name, tag=None, force=False):
```
vsoch, line 461: I’ll need to think this over - I like the separate functions but I don’t like the redundancy - that makes it harder to maintain and more likely the two become out of sync.

muffato: Yep, noticed the redundancy too. And it's not just `install` and `view_install`: there are several other places (esp. in this module: `uninstall`, `check`) where we go from a module name to the container dir/path, module dir, etc, possibly with a namespace.

shpc/main/modules/base.py
```diff
@@ -475,8 +456,47 @@ def install(
...
+        """
+        Install a module in a view.
+        The module must already be installed.
```
vsoch, line 464: Perhaps the repeated logic at the beginning of both functions could be in a common load module function? And it should return some class that transparently provides all the needed variables (that aren’t in the config, which would be one of the variables).

shpc/main/modules/base.py
vsoch: Yes! So perhaps uninstall could use this shared function to?

vsoch:
okay I started a rough idea - too tired to finish tonight but will try to soon this week!

shpc/main/modules/base.py
muffato: Yep, especially as I've got a commit (not pushed yet) to create `view_uninstall` from the current `uninstall` and mirror `install`/`view_install`

shpc/main/modules/base.py
muffato: Possibly, yes. It sounds like you've got something in mind, so I'll let you play around

vsoch: In case you missed it! https://github.com/muffato/singularity-hpc/pull/4#issue-1403058615

vsoch:
https://github.com/muffato/singularity-hpc/pull/4#issue-1403058615

shpc/main/modules/base.py
muffato: I had missed it :grin: Thanks !

vsoch: Yeah I woke up in the middle of the night like some kind of goblin... just in time for Halloween! :japanese_goblin:

muffato:
Here it is, I've split `view_uninstall` from `uninstall` like I did for `install`

shpc/main/modules/base.py
```diff
@@ -331,152 +336,94 @@ def check(self, module_name):
...
-            environment_file=self.settings.environment_file,
-        )
+        module.add_environment()
+        logger.info("Module %s was created." % module.tagged_name)
+        return module.container_path
 
-        if ":" not in name:
-            name = "%s:%s" % (name, tag.name)
-        logger.info("Module %s was created." % name)
+    def view_install(self, view_name, name, tag=None, force=False):
```
muffato, line 411: This `force` controls 1 thing: whether we allow overwriting existing symlinks

shpc/main/modules/base.py
```diff
@@ -331,152 +336,94 @@ def check(self, module_name):
...
-        # We only want to load over-rides for a tag at install time
-        config.load_override_file(tag.name)
-
-        # A view is a symlink under views_base/$view/$module
-        if view:
-            if view not in self.views:
-                logger.exit(
-                    "View %s does not exist, shpc view create %s." % (view, view)
-                )
+    def install(self, name, tag=None, force=False, **kwargs):
```
muffato, line 369: This `force` is not used anymore !

shpc/main/modules/base.py
```diff
@@ -67,29 +67,34 @@ def modulefile(self):
...
-        # We need to look for the module in all views and show to the user first
-        views_with_module = set()
+        # Only uninstall from the view
+        if view not in self.views:
+            logger.exit("View %s does not exist, cannot uninstall." % view)
+        return self.views[view].uninstall(module.module_dir)
 
-        # Only populate if the command is not directed to a view
-        if not view:
+    def uninstall(self, name, force=False):
```
muffato, line 87: The `force` here is to bypass the confirmation prompt

shpc/main/modules/base.py
```diff
@@ -67,29 +67,34 @@ def modulefile(self):
     def templatefile(self):
         return "%s.%s" % (self.container.templatefile, self.module_extension)
 
-    def uninstall(self, name, view=None, force=False):
+    def view_uninstall(self, view, name, force=False):
```
muffato, line 70: The `force` here is to bypass the confirmation prompt

muffato (commented):
I've annotated the occurrences of `force`. I'll probably tidy some of them up in #590

shpc/main/modules/base.py
```diff
@@ -67,29 +67,34 @@ def modulefile(self):
...
 
-        # folder paths do not have :
-        name = name.replace(":", os.sep)
-        module_dir = os.path.join(self.settings.module_base, name)
-        container_dir = self.container.container_dir(name)
+        # Ask before deleting anything!
+        if not force:
+            msg = name + "?"
+            if not utils.confirm_uninstall(msg, force):
+                return
```
vsoch, line 80: The following one makes sense to have the msg variable, but this one can be shortened to just:
```suggestion
            if not utils.confirm_uninstall(name + "?", force):
                return
```

shpc/main/modules/base.py
vsoch: I really appreciate these comments about force (since the meaning varies between) could we add them to the function docstrings for us of  the future to see?

shpc/main/modules/template.py
```diff
@@ -46,7 +46,7 @@ def substitute(self, template):
         For all known identifiers, substitute user specified format strings.
         """
         subs = {
-            "{|module_name|}": self.settings.module_name or "{{ parsed_name.tool }}"
+            "{|module_name|}": self.settings.module_name or "{{ config.name.tool }}"
```
vsoch, line 49: Oops I think I should have changed this back! Technically they would both work, but since we are providing parsed_name for the user (from settings.yaml -> template) it's good to be consistent.

```suggestion
            "{|module_name|}": self.settings.module_name or "{{ parsed_name.tool }}"
```

shpc/main/modules/base.py
vsoch: Huh -  I wonder why we had it here before? When you refactor upgrade it likely will make sense to have meaning here - e.g., an install with `--force` would remove stuff and re-install (if I remember you are working on this in the other PR).

vsoch (commented):
This is a heck of a lot neater - I really like it! Some small tweaks, and then given you've taken the basic changes for a spin and are happy, I'm good to merge this into main and keep working. I think your other PR #590 is going to add some really good functionality, and we can finish up there and ping Marco to see what he thinks before doing a release.

And I'm also down to keep iterating on improving the new Module class to better handle some of the logic - let me know if you see / think of something that can be improved and I'll be on it!

shpc/main/modules/base.py
muffato: Before, it was there for overwriting symlinks in views, which is now in `view_install`. Yes, it may gain a new meaning (and a new name ?) in the other PR, so I wouldn't remove it just yet

muffato:
Looks good to me ! I'll rebase #590 once this is merged.

Issue #599, opened 2022-10-13
Best way to load the Singularity module automatically?


georgiastuart:
On my systems, `singularity` is behind an lmod module and not loaded by default. I'm trying to figure out the best way to inject `load("singularity")` into the modulefile generated by shpc. It's not really a view since it needs to happen for every container. How would you recommend going about this?

georgiastuart:
I may just patch it in to the modulefile template in our install RPM, but that's a bit extreme.

vsoch:
Why not use a view and add to system_modules?

georgiastuart:
@vsoch if I set the default view as an admin, will that propagate to all users?

vsoch:
@georgiastuart a view is just another module folder (with symlinks instead of files) so if you have this added to module load (or similar, however you expose modules to users) they should be able to access. That said, I've never personally tried that out because I don't run a cluster! Pinging @marcodelapierre in case he has.

georgiastuart:
Ah, I think I understand now. The default view is just for installation so that's fine, and then we can just add the appropriate path to `$MODULEPATH`. 

However, I added the module and it's still not being loaded. Not sure why.

georgiastuart:
I've got a hunch for why it's not working. Seeing if I can fix it and then will submit a PR for y'all to check out.

georgiastuart:
Closed with #600 !

vsoch:
Holy cow number 600!! That is some kind of milestone I think! :trophy:

marcodelapierre:
Hi @georgiastuart , sorry I had missed this one!


You can also achieve pre-loading singularity by means of a setting in `settings.yml`:
```
singularity_module: 
```

PR #600, opened 2022-10-13
Fix views system module loading

georgiastuart:
I was unable to get views to appropriately load system modules on our cluster (ohpc stack, lmod module system). Upon digging around in the modulefiles, I think the following are the causes:

1. In `load_view.lua` and `load_view.tcl`:

```lua
local view_module = '.view_' .. view_name
```
The view_module should be just `.view_module`, not `.view_<modulename>`. This is consistent with the documentation and behavior of the rest of shpc.

2. In `load_view.lua`:

```lua
local view_modulefile = view_dir .. '/' .. view_module .. '.lua'
``` 
The modulefile is expected, correctly, to be `.view_module.lua`. However, the saved view modulefile is just `.view_module`. To fix, I updated `views.py` with the following:

```python
modulefile_extension = ".lua" if self.module_extension == "lua" else ""
view_module_file = os.path.join(view_dir, 
                        ".view_module%s" % modulefile_extension)
```

3. In `view_module.lua`:

Since this file gets `load`ed into the resulting modulefile, we need to omit `module` from `module load(...)`.

```lua
load("{{ module }}"){% endfor %}{% for depends in depends_on %}
```

I tested this locally (using lmod) and it seems fixed with these changes. Please let me know if I missed anything or completely misunderstood how to use views.

Signed-off-by: Georgia Stuart <georgia.stuart@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/main) (0.0.x)
+ - fix views .view_module modulefile and loading (0.1.14)
  - support for system modules, depends on, in views and editor envars (0.1.13)
  - Wrappers now supported for shell/exec/run container commands (0.1.12)
  - Update add to return container yaml (0.1.11)
--- a/shpc/main/modules/templates/includes/load_view.lua
+++ b/shpc/main/modules/templates/includes/load_view.lua
@@ -1,6 +1,6 @@
 local view_dir = string.gsub(myFileName():match("(.*[/])") or ".","/%a*/$","")
 local view_name = string.gsub(view_dir,".*/","")
-local view_module = '.view_' .. view_name
+local view_module = '.view_module'
 local view_modulefile = view_dir .. '/' .. view_module .. '.lua'
 
 if isFile(view_modulefile) then
--- a/shpc/main/modules/templates/includes/load_view.tcl
+++ b/shpc/main/modules/templates/includes/load_view.tcl
@@ -1,6 +1,6 @@
 set view_dir "[file dirname [file dirname ${ModulesCurrentModulefile}] ]"
 set view_name "[file tail ${view_dir}]"
-set view_module ".view_${view_name}"
+set view_module ".view_module"
 set view_modulefile "${view_dir}/${view_module}"
 
 if {[file exists ${view_modulefile}]} {
--- a/shpc/main/modules/templates/view_module.lua
+++ b/shpc/main/modules/templates/view_module.lua
@@ -3,5 +3,5 @@
 --
 
 {% for module in system_modules %}
-module load("{{ module }}"){% endfor %}{% for depends in depends_on %}
+load("{{ module }}"){% endfor %}{% for depends in depends_on %}
 depends_on("{{ depends }}"){% endfor %}
--- a/shpc/main/modules/views.py
+++ b/shpc/main/modules/views.py
@@ -39,13 +39,17 @@ class ViewModule:
         if not os.path.exists(view_dir):
             return
         template = self.template.load("view_module.%s" % self.module_extension)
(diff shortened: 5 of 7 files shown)


georgiastuart:
This would close #599

vsoch:
@georgiastuart I agree with the changes here and think we likely had a design bug! 

The Circle "limit" is concerning because I just don't have that much running there - I'm going to try having it run again. If it turns into an issue for a second time I might need to refactor to use a different CI service.

georgiastuart:
I'll look into why the circleci test wasn't successful this time!

vsoch:
@georgiastuart take a look at test_views.py -> test_view_components. There is a test that checks the content of files, and it looks like it's still looking for "module" and might be having an issue with the path generated. Logs here:

https://app.circleci.com/pipelines/github/singularityhub/singularity-hpc/1791/workflows/a04a9649-bd16-4a82-8bdf-e451903888f0/jobs/1087

image

And if you push changes, make sure to add a note the CHANGELOG and bump the version in shpc/version.py (and that version is tied to the line in the CHANGELOG). Thank you!

georgiastuart:
@vsoch I don't know what's up with circleci today, it just errored out fetching podman. 

I added the version bump and changelog. Once the test pass, I'll squash everything for merge once you're happy with the changes!

vsoch:
I think they might need a HugOps! :people_hugging: Let's hope we can get this one run in before our server goes out...!

Screenshot from 2022-10-13 12-34-19

I'll keep retrying to run and ping you when stuffs pass.

vsoch:
Give me a few mins I'll take a look at this test error!

georgiastuart:
@vsoch I think I solved it... we'll see what circleci says!

georgiastuart:
Gotcha, I just manually kludged it in so if you have a better solution that's preferrable!

vsoch:
See https://github.com/georgiastuart/singularity-hpc/pull/1!

And if you prefer, I can do the squash and merge (and remove the co-author that I did) to give you full credit. Also, if you are participating in Hacktoberfest, I'll add the label!

georgiastuart:
Yay, tests are passing! If you want to squash and merge, go for it @vsoch . I'm fine with coauthor :)

marcodelapierre:
Hi @georgiastuart  @vsoch, great PR! 

I have one note on point 1. at the beginning.

When views were implemented, I had suggested `view_<viewname>` instead of `view_module`, to tackle the case where multiple views are used at the same time. In this scenario, there is a risk that the `load` command in a Lua module loads the wrong `view_module` depending on the order of paths in MODULEPATH.
To tackle this risk, we had introduced view specific module names: `view_<viewname>`.

I suggest to consider reimplementing back this specific aspect.

vsoch:
I think we could @marcodelapierre ! The main issue was that I forgot, and that the convention wasn't carried forward in other parts of the code (probably because you implemented that part correctly and I missed it). If you want to open up a PR to fix it up (and @georgiastuart can test to ensure it still works, because the previous implementation did not!) I definitely think we can add it back.

georgiastuart:
Happy to test it out @marcodelapierre ! 

Though, I'm wondering how having multiple views loaded would work, considering the current workflow of "installing" an image into a particular view and loading it from there. What if a user has the same module in two loaded views, or what if the views are conflicting?

For example, in the OHPC software stack, the included lmod files disallow multiple MPI stacks from being loaded. So any two views that have different MPI stacks would conflict.

I can see the utility where having multiple views in a workflow would be useful if you need to use two shpc modules from different views, but I think it'll be trickier than renaming the `.view_module` file.

marcodelapierre:
@vsoch @georgiastuart I will aim to get on it over the next few weeks and update you accordingly.

vsoch:
Sounds good! And no rush, I had three things planned this weekend and did a total of _zero_ because I got distracted by another project... fail fail fail. :laughing:

georgiastuart:
Let me know if you want me to test anything @marcodelapierre !

marcodelapierre:
Ok, opened a PR for the proposed change from `.view_module` to `.view_<view_name>`: #614

Issue #598, opened 2022-10-13
Customise container naming


audreystott:
**Is your feature request related to a problem? Please describe.**
No, but it would remove the blocker for using currently available containers

**Describe the solution you'd like**
To be able to customise the naming convention for the singularity pull

E.g. fastqc v0.11.9 is pulled as:
```~/quay.io/biocontainers/fastqc/0.11.9--0/quay.io-biocontainers-fastqc-0.11.9--0-sha256:319b8d4eca0fc0367d192941f221f7fcd29a6b96996c63cbf8931dbb66e53348.sif```

But if the naming can be customised, it would ideally be:
```~/f/a/fastqc:0.11.9--hdfd78af_1```

This is because we have an entire biocontainers Singularity containers repo (http://datacache.galaxyproject.org/singularity/) that is read-only and that we would like to use as our containers base.

**Describe alternatives you've considered**
Possibly, scripting to divert SHPC to check for `~/f/a/fastqc:0.11.9--hdfd78af_1` if `~/quay.io/biocontainers/fastqc/0.11.9--0/quay.io-biocontainers-fastqc-0.11.9--0-sha256:319b8d4eca0fc0367d192941f221f7fcd29a6b96996c63cbf8931dbb66e53348.sif` cannot be found

**Comment by vsoch** [comment]
Have you tried installing to a view? https://singularity-hpc.readthedocs.io/en/latest/getting_started/user-guide.html#installing-modules-to-a-view The names are automatically shortened.

Also my browser doesn't let me see your galaxy page:

![image](https://user-images.githubusercontent.com/814322/195491026-4cce3306-b75c-4462-91fe-954d1a276fca.png)

so if you could write out what you are wanting for your containers, it would help me to understand! If they aren't added to shpc, we have automation means to do that (and I could certainly help!)

**Comment by audreystott** [comment]
Thanks Vanessa. 

I wasn't aware of shpc view but it looks like something we can certainly use but only after we cross the hurdle of the naming I think.

So the link I sent you is just an entire repository with biocontainers singularity containers on there. To be able to use those containers, I would like to get SHPC to see them as being present. They are all under a read-only directory `/cvmfs/singularity.galaxyproject.org`.

So for example, fastqc (version 0.11.9) are here:
```
$ ls -lh /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9*
[...]

PR #602, opened 2022-10-16
adding support to install from local container

vsoch:
This means that a user can select a registry entry (container.yaml) to use for metadata, and apply to a local container. This still requires choice of an explicit registry entry for metadata selection.

This won't close an issue explicitly, but might help with #598 


shpc/client/__init__.py
```diff
@@ -100,6 +100,7 @@ def get_parser():
     install.add_argument(
         "install_recipe",
         help="recipe to install\nshpc install python\nshpc install python:3.9.5-alpine",
+        nargs="+",
     )
```
muffato, line 104: I don't like that `nargs="+"` will take every remaining argument into the list. Maybe even the `--` options thus negating `ensure_no_extra` ?
I'd rather add an extra argument that has `nargs="?"` for the sif only, see below

```suggestion
    install.add_argument(
        "install_recipe",
        help="recipe to install\nshpc install python\nshpc install python:3.9.5-alpine",
    )
    install.add_argument(
        "container_image",
        help="path to an existing container image for this software",
        nargs="?",
    )
```

shpc/client/install.py
```diff
@@ -9,9 +9,17 @@ def main(args, parser, extra, subparser):
...
+
+    # If a second is provided, it's the container URI
+    container_image = None
+    if args.install_recipe:
+        container_image = args.install_recipe.pop(0)
+
     shpc.utils.ensure_no_extra(extra)
 
-    if args.install_recipe.startswith("gh://"):
+    if install_recipe.startswith("gh://"):
```
muffato, line 22: If the change above is accepted, no need for this

shpc/client/install.py
```diff
@@ -25,8 +33,11 @@ def main(args, parser, extra, subparser):
     cli.settings.update_params(args.config_params)
 
     # And do the install
-    cli.install(args.install_recipe, force=args.force)
+    cli.install(install_recipe, force=args.force, container_image=container_image)
```
muffato, line 36:
```suggestion
    cli.install(args.install_recipe, force=args.force, container_image=args.container_image)
```

shpc/client/install.py
```diff
@@ -25,8 +33,11 @@ def main(args, parser, extra, subparser):
...
     if cli.settings.default_view and not args.no_view:
         cli.view_install(
-            cli.settings.default_view, args.install_recipe, force=args.force
+            cli.settings.default_view,
+            install_recipe,
+            force=args.force,
+            container_image=container_image,
```
muffato, line 42:
```suggestion
            args.install_recipe,
            force=args.force,
            container_image=args.container_image,
```

muffato (commented):
Starting with some comments about the CLI

vsoch:
@muffato I think I fixed the client interactions to your liking (and I like them better too): https://github.com/singularityhub/singularity-hpc/pull/602/commits/d595c7fff9b037e386f4500999d02d35fe582175

shpc/main/modules/module.py
```diff
@@ -97,6 +86,49 @@ def container_path(self):
...
+    def add_local_container(self, container_image):
+        """
+        Set the module container to be a local container (copied over)
+        """
+        basename = os.path.basename(container_image)
+        container_dest = os.path.join(self.container_dir, basename)
+        if not os.path.exists(self.container_dir):
+            utils.mkdir_p(self.container_dir)
+        if not os.path.exists(container_dest):
+            shutil.copyfile(container_image, container_dest)
```
muffato, line 98: I can see pros and cons for both doing a copy (shpc remains in control of its containers, but it costs extra disk space) vs a symlink (no duplication on disk, but risk that the source file will go away.

I think I was expecting a symlink here ... Could this be an option ?
@audreystott : what behaviour would you expect ?

shpc/main/modules/module.py
```diff
@@ -97,6 +86,49 @@ def container_path(self):
...
+                "Expected container defined by path %s not found in %s."
+                % (self.config.path, self.config.entry.dirname)
+            )
+        container_dest = os.path.join(self.container_dir, self.config.path)
+
+        # Note that here we are *duplicating* the container, assuming we
+        # cannot use a link, and the registry won't be deleted but the
+        # module container might!
+        if not os.path.exists(container_dest):
+            shutil.copyfile(self._container_path, container_dest)
```
muffato, line 118: I'm now noticing that the copy strategy was already in use. This needs to be the same implementation as in `add_local_container`. Could we perhaps change the logic in `add_container` to:
1. If `container_image` is not set, but `self.config.path` is: set `container_image` to `os.path.join(self.config.entry.dirname, self.config.path)` (line 105)
2. Check the file referred to by `container_image` exists
3. Do the copy/symlinking

vsoch:
The preference for copy is based on the observation that people leave containers around a filesystem rather casually, and forget what they are for and/or delete them. Doing a move is unexpected because their local container would disappear, but a symlink isn’t reliable that it won’t be deleted.

My assumption here is taking the container and putting it alongside shpc so the user can’t accidentally delete it, and leaving the original for the user to handle - either keeping or deleting. It takes more space, yes, but is more conservative in terms of the risk of losing something.

However, I think we should support a use case like @audreystott - where there is some confidence the original won’t be deleted. My proposal here is to add a settings option to turn on/off creating a symlink, where the default is to be conservative and use a copy.

shpc/main/modules/module.py
vsoch: ping @audreystott - it sounds like you'd want to simply use the original path (without copying) so I'll figure out how to add a setting for that.

shpc/main/modules/module.py
vsoch: Why not just use the original path that is provided? it's assuming the user has the container already where he/she wants it.

shpc/main/modules/module.py
audreystott: Yep that’s right! All of the biocontainers are already pulled into that server and named, so SHPC won’t have to pull them.

vsoch: okay - just added support for `--keep-path` which should install the container as is. Next we need a matching of the registry containers with local files - you can likely pick one out to test locally with this branch! I'm off to bed, back tomorrow.

audreystott: Hmm. It still creates the new sif file. (I am on the correct branch and on this PR).

I did this:
```
$ shpc add /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0 fastqc
```
And the result is the same container.yaml file:
```
# change this to a URL to describe your container, or to get help
url: https://singularity-hpc.readthedocs.io

# change this to your GitHub alias (or another contact or name)
maintainer: Dinosaur

# Any custom features?
# features:
#  gpu: true

# Put custom aliases here
# aliases:
#  echo: /usr/bin/echo

# custom environment variables
# env:
#  breakfast: pancakes

# A description to describe your container
description: A custom container to do X.
latest:
  latest: sha256:56303eba371672d61c9a497702c207e4e47a9c8766d1edc3d9d75351f00332f5
tags:
  0.11.9--0: sha256:56303eba371672d61c9a497702c207e4e47a9c8766d1edc3d9d75351f00332f5
  latest: sha256:56303eba371672d61c9a497702c207e4e47a9c8766d1edc3d9d75351f00332f5
path: sha256:56303eba371672d61c9a497702c207e4e47a9c8766d1edc3d9d75351f00332f5.sif
```
And hence same result with installing:
```
$ shpc install fastqc:0.11.9--0
Error creating path /cvmfs/singularity.galaxyproject.org/all/fastqc/0.11.9--0, exiting.
```

shpc/main/modules/module.py
muffato: Even simpler :)

shpc/main/modules/module.py
vsoch: @audreystott I think you are in a different (completely opposite!) timezone, so let me know when you give the above a try! This should only be generating files in your module directory, which can be separate from where the containers are. If this works without a hitch we should be good to merge here, and tonight I'll try to finish up work to update the registry for the 8k containers, and then likely tomorrow / later this week come up with a script (or strategy) to install a latest based on the containers you find (and what is available in the registry!)

Apologies there are a lot of pieces but I haven't forgotten your use case!

audreystott: @vsoch sorry, typical user not reading the doco properly. I must be doing something wrong as it's giving the same error. 

Just to check, is it correct that I use my own registry and I do `shpc add /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0 fastqc:0.11.9--0` and then `shpc install fastqc:0.11.9--0 /cvmfs/singularity.galaxyproject.org/fastqc:0.11.9--0 --keep-path`?

```
$ shpc add /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0 fastqc:0.11.9--0
fastqc:0.11.9--0 already exists and will be updated!
Registry entry fastqc was added! Before shpc install, edit:
/home/ubuntu/shpc/registry/fastqc/container.yaml

$ shpc install fastqc:0.11.9--0 /cvmfs/singularity.galaxyproject.org/fastqc:0.11.9--0 --keep-path
Error creating path /cvmfs/singularity.galaxyproject.org/all/fastqc/0.11.9--0, exiting.
```
My timezone is UTC+8.

vsoch: No it’s not - look at the new docs here, you just want the command that I showed you, an install with a container identifier, then the SIF, then the flag. You shouldn’t be running ads at all!

vsoch: Also make sure you delete the module first I’d you accidentally installed something.

audreystott: Which registry is it supposed to look at? The quay.io/biocontainers one?

audreystott: If using the shpc-registry one, `registry: [/home/ubuntu/shpc-registry/quay.io/biocontainers]`.

Then:
```
$ shpc install fastqc:0.11.9--0 /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0 --keep-path
Error creating path /cvmfs/singularity.galaxyproject.org/all/quay.io/biocontainers/fastqc/0.11.9--0, exiting.
```

vsoch: Ah okay! So I think there must be a spot we missed where it tries to create the container directory (and we don't want that). I'm having dinner (and not wearing glasses!) but I'll take a look right after!

audreystott: :D All good. No rush!

vsoch: @audreystott could you run these commands to get the paths for your modules and containers to be installed and show them to me?

```
$ shpc config get module_base
module_base                    /home/vanessa/Desktop/Code/shpc/modules
$ shpc config get container_base
container_base                 /home/vanessa/Desktop/Code/shpc/containers
```
I'm suspicious it's trying to write to this random location with "all" - it shouldn't do that unless you've specific to write there. We will need a directory to write the modules, however, and we will need write there (and this can be separate from the containers).

audreystott: @vsoch 
```
$ shpc config get module_base
module_base                    /home/ubuntu/shpc/modules
$ shpc config get container_base
container_base                 /cvmfs/singularity.galaxyproject.org/all
```
I had it set to install to that container_base location, so that it would see that the container is already there.

vsoch: Ah, so how about unset that for now - it will default to the same as the module base (which I think is what we want).  Then again do:

```
$ shpc install fastqc:0.11.9--0 /cvmfs/singularity.galaxyproject.org/fastqc:0.11.9--0 --keep-path
```
and then see if the error still happens!

audreystott: You solved it!! The module is created and in there sits the correct path to the container!
```
Container (available through variable SINGULARITY_CONTAINER):

 - /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0
 ```

```
$ shpc list
  quay.io/biocontainers/fastqc:0.11.9--0
```

vsoch: Hooray!! So give that a good testing, try to break it, and when you are happy I’ll merge and release here. I’m almost done with an action to incrementally update our (TBA) much larger container registry, so when that is merged I’ll follow up with the container.yaml files and then get started on some means to mass install!

audreystott: Amazing! I will go and test a whole bunch of it today. Please go and rest now!!

PR #603, opened 2022-10-17
Uninstall

### LINKED ISSUE #601: shpc uninstall and tool versions
**Describe the bug**

`shpc uninstall $tool` actually only deletes the latest tag of that tool, not the other tags, while letting the user think all tags are deleted.
`shpc uninstall $tool:$version` requires the tag to still be present in the registry.
 
**To Reproduce**
```
$ shpc list
quay.io/biocontainers/samtools:1.13--h8c37831_1
quay.io/biocontainers/samtools:1.15--h3843a85_0
quay.io/biocontainers/samtools:1.14--hb421002_0

$ shpc uninstall quay.io/biocontainers/samtools
Are you sure you want to uninstall quay.io/biocontainers/samtools? (yes/no)? y
$container_base/quay.io/biocontainers/samtools and all subdirectories have been removed.
$module_base/quay.io/biocontainers/samtools and all subdirectories have been removed.

$ shpc list
quay.io/biocontainers/samtools:1.13--h8c37831_1
quay.io/biocontainers/samtools:1.14--hb421002_0
```

I also noticed that `shpc uninstall` requires the version to be present in the registry to accept it. In the example above the true version in the registry is `1.13--h8c37831_0` (ending with 0), and I've locally renamed the installed module to `1.13--h8c37831_1` to mimic an outdated version one that one may want to uninstall. `shpc uninstall` doesn't let me remove it
```
$ shpc uninstall quay.io/biocontainers/samtools:1.13--h8c37831_1
quay.io/biocontainers/samtools is not a known identifier. Choices are:
1.10--h2e538c0_3
1.11--h6270b1f_0
1.12--h9aed4be_1
1.13--h8c37831_0
1.14--hb421002_0
1.15--h3843a85_0
```

**Expected behavior**

I would expect `shpc uninstall $tool:$version` to remove a particular version (regardless of what the registry knows about it) and `shpc uninstall $tool` to remove all versions currently installed.

**Version of Singularity and Singularity Registry HPC Client**

I'm on 6b553c92884e9ef450f01353e4797ea7f7e5f147

I'm happy to fix that if you agree with the intended behaviour.

muffato:
Fixes #601 

The problem was that the `Module` object was too tightly bound to the registry. All we need a light-weight object that can provide `container_dir` and `module_dir`. I've made this the `Module` object, as simple as possible, from which a `RegistryModule` derives. Most of what used to be in `Module` is now in `RegistryModule`.

Something important to notice. This base `Module` doesn't have the concept of a tag as a separate property. A tag can still be appended to a tool name, separated by a colon, and the object will transform it into a sub-directory.


vsoch:
So conceptually, what is the difference now between `Module` and `RegistryModule` and where is `Module` alone used? 

Where my head is going is - if we don't have an actual usage of `Module`, doesn't it make sense to just have one class? :thinking:  But I might have missed a detail - about to have dinner and distracted!

muffato:
They're both used:
 - Module in uninstall functions and as a base class
 - RegistryModule in install functions

RegistryModule expands Module by getting its configuration from the Registry / container.yaml

vsoch (commented):
This is looking good! I slept on it, and my one critique is that I don't think `RegistryModule` is the right name - is it actually a module? E.g., for the `new_module` and associated `Module` we actually install into the module namespace. But the current `RegistryModule` is more akin to a Registry _entry_ - that eventually gets mapped to a module.  Is that about right?

So I think  the simplest thing to do is rename `RegistryModule` to `RegistryEntry` (and `new_registry_entry` and then (I think?) it will read more clearly when I'm reading the code. Right now when I'm reading I have to keep going back and reminding myself which is which, and this would be bad news for future me that undoubtedly won't remember!

muffato:
There is already the concept of a registry entry in [ContainerConfig](https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/container/config.py#L71-L80) although the `entry` variable is meant to be a `Result`, see [here](https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/modules/base.py#L206-L208) and [there](https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/client.py#L103-L106), and is also called "entry" [here](https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/client.py#L243-L244).

I agree `RegistryModule` isn't great. I wanted to keep the initial name (`Module`) and add in the fact that it is backed by a registry entry. I think "module" isn't that bad, as we don't already have a class for each module, and it brings the right things together: a name, the path to the container, and the path to the module directory.
Maybe there could be a single `Module` object restricted to this bare minimum, and all the rest would be offloaded to the constructor or different `new_from` methods ?
1. new_from(installed_name) is what is used in uninstall functions, and has to assume that stuff is installed where expected
2. new_from(registry_name) is what is used in install functions, and tells where stuff has to be installed.

vsoch:
What is the reason we cannot maintain one Module object, which can either be in the process of creation (RegistryModule) or assume already created (Module?)

muffato:
> What is the reason we cannot maintain one Module object, which can either be in the process of creation (RegistryModule) or assume already created (Module?)

Something quite silly. All paths rely on the `module_basepath` property, which is this implemented this way for Registry-backed modules: https://github.com/singularityhub/singularity-hpc/blob/main/shpc/main/modules/module.py#L137-L142
`self.uri` and `self.tag.name` trigger a flurry of method calls across the the ContainerConfig and many other objects.

For `shpc uninstall`, I just overrode it with whatever [name the user asked](https://github.com/muffato/singularity-hpc/blob/uninstall/shpc/main/modules/module.py#L22-L27).

That's why I was thinking that setting this `module_basepath` could be done just once in the constructor or a constructor-like function, and then indeed we would have a single object, very much like https://github.com/muffato/singularity-hpc/blob/uninstall/shpc/main/modules/module.py#L10-L44

vsoch:
Oh yes I like that idea, thumbs up 👍 from me!

muffato:
I hope you'll find the change much simpler now. I've managed to fold everything back into the constructor and the `new_module` wrapper.
It will conflict a little bit with #602, but just because of the renaming `tag_exists` → `in_registry`. That's an easy one to fix.

muffato:
(the failure on tcl, 4.7.1, docker, is just a glitch, but I can't find how to restart this job)

muffato:
I rebased on the latest `main`, and I ran `pre-commit`, to abide by the latest style rules :)

vsoch:
Will review again after dinner!

@muffato do you like the new linting? If you install as a hook it's really nice (only runs on changed files too!) You can also run on demand. I hope you like it - I really do!

shpc/main/container/config.py
```diff
@@ -225,6 +225,11 @@ def set_tag(self, tag):
         # this value will be none (and we can raise an error)
         else:
             self.tag = self.tags.get(tag)
+            if not self.tag:
```
vsoch, line 228: I think the intention of having none here is that (in other places) we can receive the value as None and then act accordingly (read the previous comment) - the best action may not always be to exit. By hard coding an exit in the function, we don't ever allow that.

shpc/main/modules/base.py
```diff
@@ -338,30 +338,29 @@ def check(self, module_name):
         at updates for entire tags. If a specific folder is provided with
         a container, check the digest.
         """
-        module = self.new_module(module_name)
+        module = self.new_module(module_name, in_registry=True)
```
vsoch, line 341: What about having a `get_module` function that basically calls `new_module` with in_registry set to True? That way, the developer reading this line isn't conflicted between the naming - the idea of a new module contrasted with already being in the registry.

shpc/main/modules/base.py
```diff
@@ -338,30 +338,29 @@ def check(self, module_name):
...
-        if ":" in name:
-            name, tag = name.split(":", 1)
-
-        module = Module(name)
-        module.config = self._load_container(module.name, tag)
-
         # Ensure the tag exists, if required, uses config.tag
-        if tag_exists:
-            module.validate_tag_exists()
+        if in_registry:
```
vsoch, line 356: For design of `get_module` - we could actually put the `in_registry` param (and checks here) entirely there, then the `get_module` function can be used on its own (with a default for True for `in_registry`. In this spot, we would basically just do:

```python
module = self.get_module(name, tag=tag, in_registry=in_registry)
```
Let me know if that doesn't make sense - it's basically adding this new function and adding some of this "in registry" logic there, since a "get" makes more sense to be checking for this over a function to "create a new one" (which we would assume not to be existing yet!)

shpc/main/modules/base.py
```diff
@@ -378,7 +377,7 @@ def install(self, name, tag=None, force=False, **kwargs):
         "force" is currently not used.
         """
         # Create a new module
-        module = self.new_module(name, tag=tag, tag_exists=True)
+        module = self.new_module(name, tag=tag, in_registry=True)
```
vsoch, line 380: This would be updated to `get_module`

shpc/main/modules/base.py
```diff
@@ -416,7 +415,7 @@ def view_install(self, view_name, name, tag=None, force=False):
         Install a module in a view. The module must already be installed.
         Set "force" to True to allow overwriting existing symlinks.
         """
-        module = self.new_module(name, tag=tag, tag_exists=True)
+        module = self.new_module(name, tag=tag, in_registry=True)
```
vsoch, line 418: Ditto Doritos.

shpc/main/modules/module.py
```diff
@@ -9,17 +9,32 @@
...
 class Module:
-    def __init__(self, name):
+    def __init__(self, config):
         """
         New module metadata and shared functions.
 
         This should be created by base.py new_module to ensure the same
         container and settings are carried forward here.
         """
-        self.name = name
```
vsoch, line 19: If config can be a config _or_ module name, let's keep the variable named to `name`. And could we visually break up the logic with a function? E.g.,

```python
class Module:
    def __init__(self, name):
        # Note for muffato: this can be set to a default of None, and overridden given it's a config

        # Move the defaults (currently below) up to the top, and they could be over-ridden by set_config
        self._uri = None

        # We can technically do this for a default, and it will be updated if we've been given a config instead!
        self.name = name

        if not isinstance(self.name, str):
            self.load_config(name)
        self.module_basepath = self.name.replace(':', os.sep)
```
It's functionally the same, but maybe a little more modular / easier to read.

shpc/main/modules/module.py
```diff
@@ -9,17 +9,32 @@
...
+                uri = str(config.name).split(":", 1)[0]
+            self.name = uri + ":" + config.tag.name
+            self._uri = uri
+            self.config = config
+
+        self.module_basepath = self.name.replace(":", os.sep)
 
         # Cache variable properties
-        self._uri = None
         self._container_dir = None
```
vsoch, line 38: These can be moved in a block to the top - the expected is that set_config could over-ride them.

shpc/main/modules/module.py
```diff
@@ -40,16 +55,6 @@ def add_environment(self):
             environment_file=self.settings.environment_file,
         )
 
-    def validate_tag_exists(self):
-        """
-        Ensure that a provided module name (and tag) exists.
-        """
-        if not self.config.tag:
```
vsoch, line 47: Let's keep this here - to follow up from the above, we want to be able to load a config that has a tag as None without it exiting. However here, we expect it to fail!

shpc/main/modules/module.py
```diff
@@ -116,28 +121,11 @@ def uri(self):
         """
         Get the uri for the module, docker / path / oras / gh
         """
-        if self._uri:
```
vsoch, line 119: Much improved to handle this logic earlier on! :raised_hands:

shpc/main/modules/module.py
```diff
@@ -116,28 +121,11 @@ def uri(self):
...
 
     @property
     def module_dir(self):
         """
         Full path to the module directory.
         """
         return os.path.join(self.settings.module_base, self.module_basepath)
-
-    @property
-    def module_basepath(self):
```
vsoch, line 139: Do you think this is better as a hard coded attribute? :thinking:

muffato:
Work in progress, and I still have one comment to address, but while doing that, I noticed something. We carry a `tag` parameters (defaulting to _None_) in `Client.install`, `Client.get_module`, `Client.view_install`, and `Client._load_container` and it's actually almost never used ! All but one entry paths provide the tag attached to the module name with a `:`, including *all* user interactions. And the [only time](https://github.com/singularityhub/singularity-hpc/blob/78f63df653109d7435c464bb1a5888d918468ac9/shpc/main/client.py#L178) we pass the tag as a separate variable, it could easily be appended to the module name.
Requesting permission to get rid of this parameter.

vsoch:
Granted!

vsoch:
I can't find the thread now, but if you want to discuss the threshold / choosing strategy for emboss I can update the containers tonight. Thanks for the help @muffato !

shpc/main/client.py
```diff
@@ -105,13 +105,15 @@ def load_registry_config(self, name):
             logger.exit("%s is not a known recipe in any registry." % name)
         return container.ContainerConfig(result)
 
-    def _load_container(self, name, tag=None):
+    def _load_container(self, name):
         """
         Given a name and an optional tag to default to, load a package
         """
         # Split name and tag
         if ":" in name:
```
vsoch, line 113: The pattern I'd prefer here is:

```python
tag = None
if ":" in name:
    ...
```
Instead of the if/else

shpc/main/modules/base.py
```diff
@@ -236,14 +236,14 @@ def docgen(self, module_name, registry=None, out=None, branch="main"):
...
+        config = self.get_module(module_name).config
 
         out = out or sys.stdout
         aliases = config.get_aliases()
         template = self.template.load("docs.md")
         registry = registry or defaults.github_url
-        github_url = "%s/blob/%s/%s/container.yaml" % (registry, branch, module_name)
-        registry_bare = registry.split(".com")[-1]
+        github_url = "%s/blob/%s/%s/container.yaml" % (registry, branch, config.name)
+        registry_bare = registry.split(".com/")[-1]
```
vsoch, line 246: Doesn't this always assume we have a .com with a trailing slash? What if we kept the same but then did a final strip of `/`?

shpc/main/modules/templates/docs.md
```diff
@@ -27,7 +27,7 @@ $ shpc install {{ name }}
 Or a specific version:
 
 ```bash
-$ shpc install {{ name }}:{{ versions.0 }}
```
vsoch, line 30: Ah! So it looks like we pass multiple attributes for config here (e.g., this example version, name and description)! Akin to how I did this for the other modules templates, let's clean up this render by just passing the config (and using it in that context in the templates). E.g.:

 - description -> config.description
 - name -> config.name
 - example_version -> config.tag.name
 - latest -> config.latest.name
etc.

And if you want we could change the current config (json dump of it) to config_json or something similar.

vsoch (commented):
This is looking great! My remaining comments are mostly nits. Also make sure to bump version and update changelog!

shpc/main/modules/templates/docs.md
muffato: Really good idea !

shpc/main/modules/base.py
muffato: Yes it assumes there is always a `/` after `.com`, because as far as I could see, `registry` is always a complete URL, even the [default value](https://github.com/singularityhub/singularity-hpc/blob/main/shpc/defaults.py#L24).
That said, and to make this bit future-proof (not all registry may be hosted on a `.com` domain), let's use `urllib` to extract the path from the URL ?

vsoch: Yes I think that's a fantastic idea - indeed custom GitLab registries don't always end in .com!

muffato:
All done !

shpc/main/modules/templates/docs.md
```diff
@@ -1,33 +1,33 @@
...
+latest: "{{ config.latest.name }}"
+container_url: "{{ config.url }}"
 {% if aliases %}aliases:{% for alias in aliases %}
  - "{{ alias.name }}"{% endfor %}{% endif %}
-versions:{% for version in versions %}
+versions:{% for version in config.tags.keys() %}
  - "{{ version }}"{% endfor %}
-{% if description %}description: "{{ description }}"{% endif %}
-config: {{ config }}
+{% if config.description %}description: "{{ config. description }}"{% endif %}
```
vsoch, line 14:
```suggestion
{% if config.description %}description: "{{ config.description }}"{% endif %}
```

vsoch (commented):
One tiny typo, then ready for merge!

vsoch (approved):

vsoch:
I'm a pancake I forgot to merge this! :laughing:

muffato:
Thank you !

PR #608, opened 2022-10-23
workflow version updates

vsoch:
Okay I'm sort of fan-girling over a new tool I made today! It's not done yet (testing and docs to do, and a few more updaters) but the basic functionality to update versions and set-output is finished, and I'm testing in the PR here. https://github.com/vsoch/action-updater. 

Here is what it looks like after a clean:

image

Since we are loading with ruamel and then saving again, this is why there is some extra formatting done as well!

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,32 +8,32 @@ jobs:
   generate-docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-        with:
-          ref: "gh-pages"
+    - uses: actions/checkout@v3
+      with:
+        ref: gh-pages
 
-      - name: Create conda environment
-        run: conda create --quiet -c conda-forge --name shpc spython
+    - name: Create conda environment
+      run: conda create --quiet -c conda-forge --name shpc spython
 
-      - name: Install shpc
-        run: |
-          export PATH="/usr/share/miniconda/bin:$PATH"
-          root=$PWD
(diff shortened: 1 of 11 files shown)


.github/workflows/docs.yml
```diff
@@ -8,32 +8,32 @@ jobs:
...
+        source activate shpc
+        /bin/bash generate.sh
 
-      - name: Deploy 🚀
-        uses: JamesIves/github-pages-deploy-action@4.1.1
-        with:
-          branch: gh-pages
-          folder: .
+    - name: Deploy 🚀
+      uses: JamesIves/github-pages-deploy-action@ba1486788b0490a235422264426c45848eac35c6 # v4.4.1
```
muffato, line 36: Just curious: why did it convert some version numbers to checksums ?

muffato: Nevermind, I found the answer at https://github.com/vsoch/action-updater#version (RTFM, I suppose ...)

vsoch: That’s the default for untrusted repos, to use a commit. There is currently a short list in settings (e.g GitHub maintained ones) that we use major tags for from releases.

vsoch: haha, I'm happy to answer your questions directly @muffato ! Those docs aren't complete or really nice looking yet :)

muffato: 👍🏼

PR #613, opened 2022-10-27
adding biocontainers example install script

vsoch:
This is an example script to install existing containers matched to modules! @audreystott the 8k+ containers are merged in the registry (!!) so this is an example (you can test) to install them. Important points:

- Please see the usage at the top - do a dry run first!
- We first try to match exactly to a tag then we fall back to the entire container name
- There will be some that don't have matches - these were containers we couldn't dump to get their aliases. And thinking about this after the fact - we might still want container.yamls for them (just without aliases). Let me know what you think -I'll need to tweak the script to ensure they get added too!

Off soon for the evening - let me know if/when you get a chance to try this out!

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,0 +1,12 @@
+# Examples
+
+This set of examples includes the following:
+
+## Biocontainers Update
+
+The intention of this script is to instantiate a client, and then match
+containers you have in some root from [Galaxy Project Depot](https://depot.galaxyproject.org/singularity/)
+on your local filesytem to install to an shpc registry (as modules). The script
+tries to be efficient and instantiate one remote registry that has all the containers.
+
+### Usage
--- a/example/biocontainer-match.py
+++ b/example/biocontainer-match.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+
+# Match biocontainers on the local filesystem to registry entries.
+# Install won't copy the container, but rather just use the provided path!
+# You should still set up your module_base before doing this (as we need
+# to write module files.
+
+# **ALWAYS** dry run first!
+#      python biocontainer-match.py --dry-run
+# At the root of your depot do:
+#      python biocontainer-match.py
+# Force module reinstall (rewrite of module file)
+#      python biocontainer-match.py --force
+# From another directory do:
+#     python biocontainer-match.py --containers /path/to/depot/root
+# If the namespace ever changes:
+#     python biocontainer-match.py --containers /path/to/depot/root --namespace quay.io/biocontainers
+# Only one namespace is currently supported!
+
+import argparse
+import os
+import re
+import sys
+
+from shpc.logger import logger
+from shpc.main import get_client
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(
+        description="SHPC BioContainer Adder",
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+    parser.add_argument(
+        "--dry-run",
+        help="Dry run only! This will preview installs.",
+        default=False,
+        action="store_true",
+    )
+    parser.add_argument(
+        "--force", help="Force reinstall of module", default=False, action="store_true"
+    )
+    parser.add_argument("--containers", help="Path to depot root.", default=os.getcwd())
+    parser.add_argument(
+        "--namespace",
+        help="Namespace to use for containers (quay.io/biocontainers)",
+        default="quay.io/biocontainers",
+    )
+    return parser
+
+
+def main():
+
+    parser = get_parser()
+
+    # If an error occurs while parsing the arguments, the interpreter will exit with value 2
+    args, extra = parser.parse_known_args()
+
+    # Show args to the user
+    print(" containers: %s" % args.containers)
+
+    if not os.path.exists(args.containers):
+        sys.exit(f"Path {args.containers} does not exist.")
+    depot = os.path.abspath(args.containers)
+    cli = get_client()
+
+    # Find all paths that match the pattern of a name:tag
+    for path in os.listdir(depot):
+        if not re.search("^(.*):(.*)$", path) or os.path.isdir(path):
+            continue
+        repo, tag = path.split(":", 1)
+        container = f"{args.namespace}/{repo}"
+        tagged = f"{container}:{tag}"
+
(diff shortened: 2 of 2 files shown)


audreystott:
Thanks for this @vsoch. As we do not have access to `cd` into the depot, I did this `python3 biocontainer-match.py --containers /cvmfs/singularity.galaxyproject.org/all --force` and got a long list of no matches. There weren't any installs/modules added. :(

vsoch:
I thought the containers were under the root of that (e.g., not all?) Can you show me an ls of the that root and the all? You need to set that `--containers` root to be where the container (SIF) files are.

audreystott:
The containers are under `all`. The root is:
```
$ ls /cvmfs/singularity.galaxyproject.org/
1  2  a  all  b  c  d  e  f  g  h  i  j  k  l  m  n  o  p  q  r  s  t  u  v  w  x  y  z
```
And to complicate things a little, which we won't use, is for e.g. fastqc:0.11.9--0:
```
$ ls -lh /cvmfs/singularity.galaxyproject.org/f/a/fastqc:0.11.9--0
lrwxrwxrwx 1 cvmfs cvmfs 26 Jan 25  2020 /cvmfs/singularity.galaxyproject.org/f/a/fastqc:0.11.9--0 -> ../../all/fastqc:0.11.9--0
```
```
$ ls -lh /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0
-rw-r--r-- 1 cvmfs cvmfs 253M Jan 25  2020 /cvmfs/singularity.galaxyproject.org/all/fastqc:0.11.9--0
```
Basically a symlink to the actual container under `all`.

vsoch:
Wait, did you clone the shpc-registry there (with the letters) or is that the organized containers from the galaxy project?

audreystott:
That's the organised containers from the Galaxy project.

vsoch:
Ah ok! So we just need to tweak the script to do a recursive find there, and parse the subdirectories (which I wasn't expecting) to parse into a container name and tag to look for in the registry. The `--containers` variable can point to the same place, but the script needs to be modified a bit.

Are you familiar with Python so this is something you'd like to try? If so - give it a shot! If not, could you please show me a tree for what the structure looks like (and what is the purpose of the letters and "all" - did you create the "all" for the shpc modules (so it's empty?)

audreystott:
I am not super familiar with Python but can try to tweak it! 

But I am not too sure what you mean with the subdirectories. We can just have the script find it in the `all` location as that is where the containers actually are. The ones subdirectories i.e. the letters are just Galaxy's way of trying to index their containers in a 'neat' way perhaps. 

It's hard to show the tree because the first level is as I showed with the `ls` above. If I do 2 levels, it will print out all the biocontainers under `all`.

vsoch:
okay so if you were to show me all, it would look like:
```
all/
   samtools:version
   another:version
```
just like the names in the online list?

audreystott:
Yep!

audreystott:
```
$ ls /cvmfs/singularity.galaxyproject.org/all | wc -l
79281
```
:O

vsoch:
okay! So I think this could be a good learning opportunity then. The way I debug Python is with an Ipython embed statement, which is going to let me embed into an interactive terminal to test things out. So (since I can't be there to do this on your filesystem) what you want to do is source a Python environment and make sure you have IPython, e.g.,

```bash
$ pip install IPython
```
And then before these lines: https://github.com/singularityhub/singularity-hpc/pull/613/files#diff-668e8b085ee99a6af7f4bc8608464dcd42d87d28928753735ccc92df994eafa7R67-R78

put:

```
import IPython
IPython.embed()
sys.exit()
```
I always put the exit at the end so it exits when I'm done (without continuing to run). Then what you want to do is step through that block by copy pasting bits of logic - and do sanity checks as you go. E.g.,, some suggested checks:

- The args.containers should point to the correct location
- os.listdir of that should not be empty
- Choose a container in the list (e.g., os.listdir(<path>)[10] and then test to see if the regular expression returns a result (it should).
- Sanity check what we get for repo, tagged, container, etc.

What we basically want to do here is understand _why_ the script is telling us there aren't any matches. I can almost guarantee it's a small detail or oversight, something about a path or symbolic link or something we didn't expect! When we know why the matches aren't happening, then we can work on a fix. Sound good?

vsoch:
And I'm off to bed soon - but if you can't figure out the bug, then can you give me some detail about how to reproduce (a small subset) of your setup, and I can take a shot tomorrow? I tested this by downloading a few of those containers manually into a directory, pointing the script to it, and it worked for me, so I suspect I missed a detail.

audreystott:
Ok I’ll give this a try in about an hour! Have done something similar with other Python code recently. Don’t wait up for me.

vsoch:
and I see one tiny bug (going to fix now) so we don't install repeats!

audreystott:
Ok so I tried it and it looks like there are two aspects:
1. the namespace `quay.io/biocontainers` is not a prefix of the containers in the repo, but
2. even omitting the `{args.namespace}` for the container definition, the output is the same.
Where it starts to not show is at this stanza:
```
match = cli.registry.find(tagged)
    if not match:
        tagged = container
        match = cli.registry.find(tagged)
```
`print(containers)` and `print(tagged)` gave all the containers and containers with tags. (I couldn't/didn't know how to do just a subset of the list. `)

If it helps to debug, I am happy for you to have access to a VM that you can use /cvmfs/singularity.galaxyproject.org/. Just let me know your pub key.

audreystott:
Ah Vanessa! Sorry. I found out what's wrong and it's not to do with your script at all!!! It's because I had my `container_base` in my `settings.yml` file empty - for when trialing the `shpc install` command previously! Now that I have set it to `/cvmfs/singularity.galaxyproject.org/all` it now returns a list of "would be installing" and "not match"s for the dry-run!

vsoch:
Are you sure that's it? I was going to suggest that the match needs to be taken on os.path.basename(path) before processing it into the repo and tag.

audreystott:
Yes definitely as I have just done the same on the new VM I spun up for you. I figured that might have been the case as you imported the shpc-client for the script, which I figure has to do with it all? :D (VM address sent to you via email).

Issue #565, opened 2022-07-18
Generating shell scripts instead of functions for wrappers without specifying on the container.yaml


surak:
As the wrapper_scripts option is now enabled by default, shpc will only create scripts in cases the container.yaml file has the docker_scripts/singularity_scripts tag.

My suggestion would be that, instead of creating functions for calling the containers, simple default shell scripts with such content be created instead. This way, `slurm` and others will be happy to call it. 

The current default is to only generate shell scripts if there's enough info on the container.yaml.

Changes needed:
- The module would need to export a path, which is not done now.
- Each function (like container_run, container_exec, container_container, container_shell) gets created as a shell script instead of a function.

What do you think? That's better than to have to add a script/wrapper entry to every single container in the registry.

vsoch:
Hmm this sounds like a bug - the scripts should generate using the default templates from the settings.yaml. @muffato was this added with the PR that updated wrapper scripts?

muffato:
Yes, it sounds like a bug too ! Scripts should be created regardless of whether they have a template defined / overwritten in `singularity_scripts`. I can take a look later tonight

vsoch:
@surak that would be great if you could give the complete steps you took so we can reproduce! I can try as well.

surak:
```bash
 find .
.
./nvcr.io
./nvcr.io/nvidia
./nvcr.io/nvidia/tensorflow
./nvcr.io/nvidia/tensorflow/.version
./nvcr.io/nvidia/tensorflow/22.06-tf2-py3
./nvcr.io/nvidia/tensorflow/22.06-tf2-py3/module.lua
./nvcr.io/nvidia/tensorflow/22.06-tf2-py3/99-shpc.sh
./nvcr.io/nvidia/tensorflow/22.06-tf2-py3/nvcr.io-nvidia-tensorflow-22.06-tf2-py3-sha256:ea76aabaa17ee0e308a443fa1e7666f98350865f266b02610a60791bd0feeb40.sif
```

I can confirm that quay.io works with the aliases containing in the container.yaml file, but not the template functions.

surak:
So, I think it's a different thing.

In quay.io/biocontainers/blast/container.yaml, the aliases are like this:

```
aliases:
  blast_formatter: /usr/local/bin/blast_formatter
  blastdb_aliastool: /usr/local/bin/blastdb_aliastool
  blastdbcheck: /usr/local/bin/blastdbcheck
[...]

PR #628, opened 2023-02-03
modulefiles templates: shell aliases always defined

marcodelapierre:
Signed-off-by: Marco De La Pierre <marco.delapierre@gmail.com>

### DIFF
--- a/shpc/main/modules/templates/docker.lua
+++ b/shpc/main/modules/templates/docker.lua
@@ -76,7 +76,7 @@ if (myShellName() == "bash") then
 {% endfor %}
 end{% endif %}
 
-{% if wrapper_scripts %}{% else %}set_shell_function("{|module_name|}-container", "echo " .. containerPath, "echo " .. containerPath)
+set_shell_function("{|module_name|}-container", "echo " .. containerPath, "echo " .. containerPath)
 
 -- set_shell_function takes bashStr and cshStr
 set_shell_function("{|module_name|}-shell", shellCmd,  shellCmd)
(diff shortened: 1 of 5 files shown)


marcodelapierre:
Hi @vsoch I have noticed this today, as I was testing the latest release `0.1.18` vs `0.0.57`, which we have been using so far on our supercomputer.
I am proposing to always generate the shell aliases for things like `-shell`, `-container` and so on, even when we use wrapper scripts.
Did you remove them for some specific issue or bug?

From my perspective, I understand It is a tiny inconsistency, but it should not generate big issues; our rationale was that these aliases are by nature mostly used interactively anyway, so should not pose any issue.
On the contrary, not having them can be quite annoying, as it takes out functionality; in particular `-shell` and `-container` can be quite often used. 
In summary, I think reinstating them is the lesser evil.

Eventually, in the future we might provide wrapper scripts for these as well, to be used when wrappers are turned on for the main application commands.

What do you think?

vsoch:
@marcodelapierre I think if wrapper scripts are generated, we generate the aliases but as scripts instead.  The last time we chat about it was for this issue https://github.com/singularityhub/singularity-hpc/issues/565 and I followed up with https://github.com/singularityhub/singularity-hpc/pull/586 that ensured they were generated, e.g., here: https://github.com/singularityhub/singularity-hpc/blob/ee4881a1c505ff3c0d21dbe57e547b1b791d4932/shpc/main/wrappers/__init__.py#L36

So if they aren't in the bin of the module directory, that's probably a bug.

marcodelapierre:
Aaaaargh!! It is really the sign that my weekend is looming...!!

I even tested `-shell`, then forgot, did not find it in the module file, branch out, open the PR ... 🙈
All good, no bug, apart from the weekend bug..!!

Thanks for your patience ;-)

Issue #630, opened 2023-02-14
Support remote registries on self-hosted gitlab instances


phue:
Hi @vsoch,

First of all, thanks for this great tool!
I'm trying to deploy a custom registry on a self-hosted Gitlab instance and use it on our HPC system.
However, it seems that the url to gitlab pages currently works with the gitlab.io domain only.

https://github.com/singularityhub/singularity-hpc/blob/68a000507e42f77b051b0b4a6ca70b230d53d5c5/shpc/main/registry/remote.py#L110

In my specific case, I would need this pattern: `https://%s.pages.%s.rlp.net/%s/library.json` instead.
Would it be possible to add a config option for a custom url?

vsoch:
Oh we could definitely handle that! Perhaps we should have a registry options with the url format string there?

vsoch:
But perhaps it could be derived from the source_url instead? Can you show me your source URL?

phue:
> But perhaps it could be derived from the source_url instead? Can you show me your source URL?

I didn't think of that, that should work.

Essentially, the `source_url` for selfhosted instances is `gitlab.example.com/<username>/<reponame>` which gets served at `<username>.pages.gitlab.example.com/<reponame>` (see https://docs.gitlab.com/ee/administration/pages/#prerequisites)

The attached pull request should do the trick, while keeping the functionality for gitlab.com and github.com

PR #631, opened 2023-02-14
add support for self-hosted gitlab instances

phue:
closes https://github.com/singularityhub/singularity-hpc/issues/630

### DIFF
--- a/shpc/main/registry/remote.py
+++ b/shpc/main/registry/remote.py
@@ -107,7 +107,7 @@ class VersionControl(Provider):
         Retrieve the web url, either pages or (eventually) custom.
         """
         domain, namespace, repo = self.source_url.split("/")[2:]
-        domain_parts = domain.split(".")
+        domain_parts = domain.split('@')[-1].split(".")
 
         if len(domain_parts) == 2:
             domain_parts[1] = "io"
--- a/shpc/main/registry/remote.py
+++ b/shpc/main/registry/remote.py
@@ -117,7 +117,7 @@ class VersionControl(Provider):
         return "https://%s.%s/%s/library.json" % (
             namespace,
             ".".join(domain_parts),
-            "/".join(repo),
+            repo,
         )
 
     def exists(self, name):
--- a/shpc/main/registry/remote.py
+++ b/shpc/main/registry/remote.py
@@ -106,18 +106,18 @@ class VersionControl(Provider):
         """
         Retrieve the web url, either pages or (eventually) custom.
         """
-        parts = self.source_url.split("/")[2:]
-        domain = parts[0].split(".")
+        domain, namespace, repo = self.source_url.split("/")[2:]
+        domain_parts = domain.split(".")
 
-        if domain == [self.provider_name, 'com']:
-            domain[1] = "io"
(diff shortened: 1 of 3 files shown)


shpc/main/registry/remote.py
```diff
@@ -106,11 +106,18 @@ def web_url(self):
         """
         Retrieve the web url, either pages or (eventually) custom.
         """
-        parts = self.source_url.split("/")[3:]
-        return "https://%s.%s.io/%s/library.json" % (
-            parts[0],
-            self.provider_name,
-            "/".join(parts[1:]),
+        parts = self.source_url.split("/")[2:]
+        domain = parts[0].split(".")
```
vsoch, line 110: This is splitting on an unknown number of `.` are you sure you don't want to be specific? 

I think what I would like to see here is parsing the url, and then formally splitting into prefix/suffix, and then using those specifically. The check for if domain == a list is also erroneous and makes assumptions about that structure.

Pinging @muffato for his thoughts since he wrote the original logic.

phue: My understanding was that the `parts` split already yields the suffix, gets rid of the `https://` prefix (and what follows is the full domain which can have a variable number of dots)

phue: Ok, I changed it to make the parsing more specific.
About the if clause: it is meant to handle the case for git{hub,lab}.com, I think it should safe to make that assumption about the structure there.

vsoch: oh definitely this is much more clear! Have you tested it with your registry and you are happy with it?

If so it is good for me - for the final tweaks please add a note to the CHANGELOG.md with a version bump, and do the same bump in the shpc/version.py file. And after that we will just want @muffato to review and approve too.

Thanks for fixing this!

muffato (approved):
Thanks @phue : this looks so simple and yet does the job !
My own start on this (#585) went down the rabbit hole. I'm happy to get this PR in, and I'll come back to mine later

shpc/main/registry/remote.py
phue: > oh definitely this is much more clear! Have you tested it with your registry and you are happy with it?
works like a charm :rocket: 

> If so it is good for me - for the final tweaks please add a note to the CHANGELOG.md with a version bump, and do the same bump in the shpc/version.py file. And after that we will just want @muffato to review and approve too.

Done :)

vsoch:
I've seen this failure before https://github.com/singularityhub/singularity-hpc/actions/runs/4182907333/jobs/7247353378 and I tried to debug it before, I'm not sure I see why the test no longer is working (e.g., it's grabbing the wrong version based on the setting).

phue:
Looks like the singularity pull fails for some reason, could be just some quirk on the runner
https://github.com/singularityhub/singularity-hpc/actions/runs/4182907333/jobs/7247353378#step:6:223

phue:
snuck in a quick change to handle repositories that need authentication credentials (they have this structure: `https://<user>:<password>@example.com/<namespace>/<reponame>`)

vsoch:
Yep the test looks like some kind of race condition, but it passed this time so we are good!

PR #636, opened 2023-03-14
Doc update

muffato:
Hi !

Another doc change, since we've got some automation in place.

Also, I would like to add somewhere that `shpc add` relies on the crane service being able to query the containers, which may fail in some occasions. I'm not sure whether to put that to the user guide or developer guide. Any thoughts ?

For instance it wouldn't work for privately-hosted containers, this is what we get in the yaml file
```
crane digest gitlab-registry.internal.sanger.ac.uk/tol-it/software/bio-interproscanwrapper:5.61-93.0-ext:
    Get https://gitlab-registry.internal.sanger.ac.uk/v2/: dial tcp: lookup gitlab-registry.internal.sanger.ac.uk
    on 169.254.169.254:53: no such host
```
And I think there was another error sometimes because crane returns the top 50 tags or so, which may not include the specific tag the user may ask

### DIFF
--- a/docs/getting_started/developer-guide.rst
+++ b/docs/getting_started/developer-guide.rst
@@ -1087,7 +1087,14 @@ without any customization for executable discovery:
     Registry entry vanessa/salad was added! Before shpc install, edit:
     /tmp/my-registry/vanessa/salad/container.yaml
 
-You could then edit that file to your liking. If you want to pull the container
+You could then edit that file to your liking.
+
+Like for ``shpc update`` :ref:`getting_started-commands-update`, tags are automatically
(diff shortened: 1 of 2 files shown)


vsoch (approved):

vsoch:
That would work! We do mention crane for `shpc update` https://github.com/singularityhub/singularity-hpc/blob/0a063533cce05f3ca86d84c486087f6b08fffec3/docs/getting_started/user-guide.rst#update and that's what we use for add https://github.com/singularityhub/singularity-hpc/blob/0a063533cce05f3ca86d84c486087f6b08fffec3/shpc/main/container/singularity.py#L160.

muffato:
Thank you for the pointer, I've added another mention in the developer guide.

PR #638, opened 2023-04-01
Add/example google cloud storage

### LINKED ISSUE #605: GCP buckets as registry
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
We would like to use GCP buckets as a registry for our containers and use those with shpc and I believe currently that is not possible. It would be really helpful for us if that capability was added. Thank you for your consideration.
**Describe alternatives you've considered**

vsoch:
This will close #605 

Note that the GCP bucket is not a registry, it's a cache for all containers / module files to be mounted somewhere.


vsoch:
@evalanguaria that should be an easy change - before you run the script, just edit your shpc settings.yaml config and ensure the containers/modules are installed to different locations. You will then need to update the script to handle the change in location (instead of just root you will have two!)

This is the simplest implementation that I would like to provide for the example for other users, so I think it's best that if you'd like it customized you are free to do something else.

evalanguaria:
@vsoch yes that makes sense. I will make the necessary modifications. Thanks a lot again for everything.

PR #652, opened 2023-06-24
Improves documentation covering the creation of recipes for local containers.

dipietrantonio:
Make it clear that the container image is bundled with the recipe when adding a local container.

This pull request addresses https://github.com/singularityhub/shpc-registry/issues/113

### DIFF
--- a/docs/getting_started/user-guide.rst
+++ b/docs/getting_started/user-guide.rst
@@ -1876,7 +1876,8 @@ and tell shpc that we want it under the ``dinosaur/salad`` namespace.
     Registry entry dinosaur/salad:latest was added! Before shpc install, edit:
     /home/vanessa/Desktop/Code/shpc/registry/dinosaur/salad/container.yaml
 
-At this point, you should open up the container.yaml generated and edit to your liking.
+The container image will be copied within the newly-created directory structure, alongside the `container.yaml` file.
+At this point, you should open up the generated `container.yaml` file and edit to your liking.
 This usually means updating the description, maintainer, aliases, and possibly providing a url
 to find more information or support. Also notice we've provided the tag to be latest. If you update this registry
 entry in the future with a new version, you'll want to provide a new tag. If you provide


docs/getting_started/user-guide.rst
```diff
@@ -1859,7 +1859,8 @@ and tell shpc that we want it under the ``dinosaur/salad`` namespace.
     Registry entry dinosaur/salad:latest was added! Before shpc install, edit:
     /home/vanessa/Desktop/Code/shpc/registry/dinosaur/salad/container.yaml
 
-At this point, you should open up the container.yaml generated and edit to your liking.
+The container image will be copied within the newly-created directory structure, alongside the `container.yaml` file.
```
vsoch, line 1862: More specifically, it will be put in `container_base` that could be a custom path specified by the user https://github.com/singularityhub/singularity-hpc/blob/2a2e30cb50a8db6fb62f25adb84074d8325ade3c/shpc/settings.yml#L43. Also, if they create a module with a local path defined (example [here](https://github.com/singularityhub/singularity-hpc/blob/2a2e30cb50a8db6fb62f25adb84074d8325ade3c/example/biocontainer-match.py#L143)) the container will not be copied, but rather preserved in that path (and referenced as such in the container.yaml). We do this for use cases where you have a huge archive of existing containers, and just want to make separate modules for them.

dipietrantonio: @vsoch thanks for the explanation, I like these features. When you say

> if they create a module with a local path defined

You mean creating the `container.yaml` manually, and NOT using the `shpc add ... ` command. Is that correct? Because using the `shpc add` command creates a copy by default?

vsoch: No, the example I linked is using the same logic as shpc add, but from within Python to specify a specific container path to write into the container.yaml recipe and not do a copy.

dipietrantonio: So this is something you can do programmatically using SHPC as a Python module, but is there a way to enable `keep_path`  through the `shpc` command line tool?

vsoch: Not currently - it’s just provided as that example right now.

dipietrantonio: Then I think it might be better not to mention this feature under the `add` command documentation, as it would confuse readers.

dipietrantonio: I also want to make sure that I got the following right:
> More specifically, it will be put in container_base that could be a custom path specified by the user
This happens when we install a container module (`shpc install`), I do not recall this happening when executing `shpc add`. When I execute `shpc add`, the container image is copied alongside the `container.yaml`, even when the `container_base` points somewhere else.

vsoch: shpc install is installing an existing container.yaml, and add is generating it.

And I disagree about not mentioning it - we should describe exactly what is possible. E.g., say that add from the command line will by default copy the container binary, and for those interested in preserving a path they can see (this example) with the link.

dipietrantonio: > And I disagree about not mentioning it - we should describe exactly what is possible
Fair enough, I will incorporate that information.

> shpc install is installing an existing container.yaml, and add is generating it.

I get that, but it still does not address my observation. Consider the following sequence of commands I have just executed (shown below). `shpc add` will copy the container in the same directory as `container.yaml`, NOT in `container_base` as you stated in the initial comment of this thread. I am being pedantic but I want to make sure that what I write in the paragraph is correct.
```
$ shpc config get container_base
container_base                 /scratch/pawsey0001/cdipietrantonio/cbase
$ # Initially, TensorFlow is not there
$ ls /scratch/pawsey0001/cdipietrantonio/cbase/
amdih  pawsey
$ ls /scratch/pawsey0001/cdipietrantonio/cbase/amdih/
gromacs/ pytorch/ 
$ shpc add tensorflow2.sif amdih/tensorflow:latest
Registry entry amdih/tensorflow was added! Before shpc install, edit:
/software/projects/pawsey0001/cdipietrantonio/setonix/python/lib/python3.9/site-packages/registry/amdih/tensorflow/container.yaml
$ ls /software/projects/pawsey0001/cdipietrantonio/setonix/python/lib/python3.9/site-packages/registry/amdih/tensorflow
container.yaml	sha256:aa8a6fd70ecc08283d9a19d62212f1a6762ef4976f3048177d47e9c543a00768.sif
$  ls /scratch/pawsey0001/cdipietrantonio/cbase/amdih/
gromacs  pytorch

```

vsoch: Ah that's a bug! Because of the lines here:

https://github.com/singularityhub/singularity-hpc/blob/2a2e30cb50a8db6fb62f25adb84074d8325ade3c/shpc/main/container/singularity.py#L99-L100

That should not just be taking the parent directory, but also accounting for the settings container base. Would you like to fix it here or would you like me to open a separate PR?

vsoch: I haven't tested closely, but here is something to try! https://github.com/singularityhub/singularity-hpc/pull/656 I both exposed keep_path to add, and also enforced using container_base.

dipietrantonio: That is great, looks correct to me. When will it be merged? So I can update this PR accordingly.

vsoch: @dipietrantonio I was hoping you might test out the two changes?

- add with a custom container path
- fixing the bug to honor container_base

When you've confirmed it works as expected we can merge.

dipietrantonio: I would suggest merging this PR as it actually describes what happens. Then change the documentation again when you fix the bug.

vsoch: I'm fine with that, thanks!

PR #656, opened 2023-07-01
fix bug with container add

vsoch:
and expose keep_path to add

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/main) (0.0.x)
+ - Fix bug with container add not respecting container_base (0.1.24)
  - Labels with newlines need additional parsing (0.1.23)
  - Do not write directly to output with shpc show (0.1.22)
  - Podman template bug (0.1.21)
--- a/shpc/client/__init__.py
+++ b/shpc/client/__init__.py
@@ -103,14 +103,6 @@ def get_parser():
         help="path to an existing container image for this software",
         nargs="?",
     )
-
-    install.add_argument(
-        "--keep-path",
-        help="if installing a local container, do not copy the container - use the provided path.",
-        default=False,
-        action="store_true",
-    )
-
     install.add_argument(
         "--no-view",
         dest="no_view",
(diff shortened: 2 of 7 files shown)


dipietrantonio:
Hi @vsoch , I did some experiments

```
$shpc add /scratch/pawsey0001/cdipietrantonio/tensorflow.sif amdih:tensorflow:latest
SINGULARITY is not installed, functionality might be limited.
Registry entry amdih was added! Before shpc install, edit:
/software/projects/pawsey0001/cdipietrantonio/setonix/python/lib/python3.9/site-packages/singularity_hpc-0.1.24-py3.9.egg/registry/amdih/container.yaml
$ ls /software/projects/pawsey0001/cdipietrantonio/setonix/python/lib/python3.9/site-packages/singularity_hpc-0.1.24-py3.9.egg/registry/amdih/
container.yaml
$ ls /scratch/pawsey0001/cdipietrantonio/cbase/
amdih  pawsey  sha256:153ff47d5dfac0807504b5c29c7ecd7d9ea6b7736ceb0d5876a1b0c1a980e0c1.sif
```
So as you can see the container is not copied anymore into the same directory of container.yaml, but directly under container_base, as indeed the code instructs to do.

I was wondering whether it would be better to recreate a directory structure for the container, containing registry and software name. In our case `container_base/amdih/tensorflow`.

Also, is the `path` field in `container.yaml` relative to `container_base` ?

vsoch:
Looks like we have some bugs! I’ll take a look tomorrow. Thanks for testing it out!

vsoch:
hey @dipietrantonio good morning! I just pushed (hopefully) a fix so that the container path includes `container_base` and the module name (that was a bug) and want to answer your question.

> Also, is the path field in container.yaml relative to container_base ?

The container.yaml is part of the registry, so this is derived from a filesystem registry path, e.g., `self.settings.filesystem_registry`

vsoch:
There is still a bug - will work on it later today after work!

dipietrantonio:
Hi @vsoch, how is it going on this PR?

vsoch:
I'm trying to remember what the bug is - looking at what I was doing, I was changing container add to have the container go under the container_base. But this is in correct - a container add isn't an install, and should be added alongside the container.yaml config (or not moved at all) since the user isn't actually installing it.

vsoch:
So I'm tempted to say this is not a bug - a container add should not be installing to the containers base, because it is simply just creating a container.yaml that then should be installed.

PR #657, opened 2023-07-03
Set ruamel.yaml version to 0.17.21

audreystott:
For shpc version 0.1.22, ruamel.yaml 0.17.32 (default latest) does not work. 0.17.21 has been tested and confirmed to work with shpc 0.1.22.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,12 +25,16 @@ def get_reqs(lookup=None, key="INSTALL_REQUIRES"):
         module_meta = module[1]
         if "exact_version" in module_meta:
             dependency = "%s==%s" % (module_name, module_meta["exact_version"])
-        else:
-            if "min_version" in module_meta:
-                dependency = "%s>=%s" % (module_name, module_meta["min_version"])
+        elif "min_version" in module_meta:
+            if module_meta["min_version"] is None:
+                dependency = module_name
             else:
-                if "max_version" in module_meta:
-                    dependency = "%s<=%s" % (module_name, module_meta["max_version"])
+                dependency = "%s>=%s" % (module_name, module_meta["min_version"])
+        elif "max_version" in module_meta:
+            if module_meta["max_version"] is None:
+                dependency = module_name
+            else:
+                dependency = "%s<=%s" % (module_name, module_meta["max_version"])
         install_requires.append(dependency)
     return install_requires
 
--- a/setup.py
+++ b/setup.py
@@ -25,11 +25,12 @@ def get_reqs(lookup=None, key="INSTALL_REQUIRES"):
         module_meta = module[1]
         if "exact_version" in module_meta:
             dependency = "%s==%s" % (module_name, module_meta["exact_version"])
-        elif "min_version" in module_meta:
-            if module_meta["min_version"] is None:
-                dependency = module_name
-            else:
+        else:
+            if "min_version" in module_meta:
                 dependency = "%s>=%s" % (module_name, module_meta["min_version"])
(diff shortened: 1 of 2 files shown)


vsoch:
Do you want to do `exact_version` or `max_version`? I think lower versions could work too?

vsoch:
Note we also need to add the `max_version` logic here: https://github.com/singularityhub/singularity-hpc/blob/2a2e30cb50a8db6fb62f25adb84074d8325ade3c/setup.py#L26-L28 (haven't had one before! ) It should look like min_version but with the sign reversed.

audreystott:
Ok let me test it

audreystott:
Ok ready for your review.

setup.py
```diff
@@ -25,11 +25,12 @@ def get_reqs(lookup=None, key="INSTALL_REQUIRES"):
         module_meta = module[1]
         if "exact_version" in module_meta:
             dependency = "%s==%s" % (module_name, module_meta["exact_version"])
-        elif "min_version" in module_meta:
-            if module_meta["min_version"] is None:
-                dependency = module_name
-            else:
+        else:
```
vsoch, line 28: It looks like the nesting of if else might be off here?

vsoch: This should be what you want, an elif

```python
# Read in requirements
def get_reqs(lookup=None, key="INSTALL_REQUIRES"):
    """get requirements, mean reading in requirements and versions from
    the lookup obtained with get_lookup"""

    if lookup is None:
        lookup = get_lookup()

    install_requires = []
    for module in lookup[key]:
        module_name = module[0]
        module_meta = module[1]
        if "exact_version" in module_meta:
            dependency = "%s==%s" % (module_name, module_meta["exact_version"])
        elif "min_version" in module_meta:
            if module_meta["min_version"] is None:
                dependency = module_name
            else:
                dependency = "%s>=%s" % (module_name, module_meta["min_version"])
        elif "max_version" in module_meta:
            if module_meta["max_version"] is None:
                dependency = module_name
            else:
                dependency = "%s<=%s" % (module_name, module_meta["max_version"])
        install_requires.append(dependency)
    return install_requires
```

audreystott: Interesting, it works for setting up SHPC. I had a similar code to yours before I then went reducing the lines down to just an if else if else. I am not familiar with elif logic - but it works!

vsoch:
Perfecto! Two tiny tweaks then we are good for merge (and these are required for all PRs, this one and into the future)

- Bump the version in shpc/version.py
- Add a corresponding note about your change in CHANGELOG.md

vsoch:
We can also skip the version bump, but you'll need to install from main. Let me know your preference!

audreystott:
Sorry I don't quite understand. Will you be re-releasing 0.1.22?

vsoch:
No, whenever we change shpc we do a new release. That is dictated by the version string in shpc/version.py (which we need to increment by 1) and the changelog tells future us what was done.

audreystott:
Right! That simplifies things a little with versioning. I think let's not worry about a version bump.

vsoch:
@audreystott I think we will need to update here - the conda build is broken because of this change: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=783260&view=logs&jobId=656edd35-690f-5c53-9ba3-09c10d0bea97&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3.

PR #660, opened 2023-09-21
Adding wrapper_base to allowed_envars

theklaloizou:
After the addition of wrapper_base option in settings.yml file when using an environment variable in the wrapper_base path the variable is not expanded correctly.

This is a PR to fix this issue by just adding wrapper_base in the allowed_envars.

### DIFF
--- a/shpc/defaults.py
+++ b/shpc/defaults.py
@@ -18,7 +18,7 @@ user_settings_file = os.path.join(
 )
 
 # variables in settings that allow environment variable expansion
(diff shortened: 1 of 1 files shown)


vsoch:
This looks good! Since you are new I won't ask for extra work (e.g., release stuffs mostly) - we can do a release for this the next time around. I think @marcodelapierre is going to do some work in a few weeks so it will be grouped with that. Thanks @theklaloizou !

vsoch (approved):

Issue #678, opened 2024-12-25
Module name in shpc configuration


Amjadhpc:
Hello i am new to Singularity hpc and trying to build containers from local file registry  singularity containers

My question is in naming of module file
For example if i have  container of application velocyto with version 0.17.13 , the following module file is created
velocyto/0.17.13/module <if module_sys is lua>
velocyto/0.17.13/module.tcl <if module_sys is tcl>

Is it possible in shpc config that the version number is the module file  instead having module. or module.tcl file
velocyto/0.17.13 <if module_sys is lua>
velocyto/0.17.13.tcl <if module_sys is tcl>

The reason behind this question is currently we are making manual module files using this method where version number is module file and we want to automate the method with shpc . We can look doing some scripts  to do this change , but were wondering if we can modify the shpc config to have it set by default.
I hope i have been able to articulate my question

vsoch:
@Amjadhpc that is a very reasonable feature request! Are you able to develop in Python or would you need help? You'd want to add a setting to settings.yaml that is something akin to `version_naming` and then a boolean that defaults to false (to preserve current behavior) and if true, you'd want to tweak the install path to instead of creating a directory with the version, to have it write a file with the module extension to that path instead. Let me know if you would like to give that a try and open a PR. Happy Holidays!

Amjadhpc:
@vsoch . Thanks.  I have attempted  what you had explained and opened a PR request .

muffato:
Hi. Alternatively, shpc already has the concept of "[views](https://singularity-hpc.readthedocs.io/en/latest/getting_started/user-guide.html#views)", which were implemented exactly to provide a module tree of the form `<name>/<version>.tcl`

Amjadhpc:
Closing the issue after the PR  merge

PR #679, opened 2025-01-01
Introducing version_naming property that makes modulefile as version …

Amjadhpc:
This is an attempt to fix issue number 26 raised by me.
I have not tested all settings

### DIFF
--- a/shpc/main/modules/base.py
+++ b/shpc/main/modules/base.py
@@ -474,10 +474,9 @@ class ModuleBase(BaseClient):
         # Get the template based on the module and container type
         template = self.template.load(self.templatefile)
         if self.settings.version_naming:
-            self._modulefile = name.split(":")[1]
             module._module_dir = os.path.dirname(module.module_dir)
             module_path = os.path.join(
-                os.path.dirname(module.module_dir), self._modulefile
+                os.path.dirname(module.module_dir), name.split(":")[1]
             )
         else:
             module_path = os.path.join(module.module_dir, self.modulefile)
--- a/shpc/main/modules/base.py
+++ b/shpc/main/modules/base.py
@@ -474,7 +474,6 @@ class ModuleBase(BaseClient):
         # Get the template based on the module and container type
         template = self.template.load(self.templatefile)
         if self.settings.version_naming:
-            module._module_dir = os.path.dirname(module.module_dir)
             module_path = os.path.join(
                 os.path.dirname(module.module_dir), name.split(":")[1]
             )
--- a/shpc/main/modules/base.py
+++ b/shpc/main/modules/base.py
@@ -414,7 +414,7 @@ class ModuleBase(BaseClient):
         name = self.add_namespace(name)
 
         module = Module(name)
-
+         
         # Pass on container and settings
         module.container = self.container
         module.settings = self.settings
@@ -450,7 +450,7 @@ class ModuleBase(BaseClient):
         "force" is currently not used.
         """
         # Create a new module
-        module = kwargs.get("module") or self.get_module(
+         module = kwargs.get("module") or self.get_module(
             name, container_image=container_image, keep_path=keep_path
         )
 
@@ -458,7 +458,10 @@ class ModuleBase(BaseClient):
         module.load_override_file()
 
         # Create the module and container directory
-        utils.mkdirp([module.module_dir, module.container_dir])
+        if self.settings.version_naming: 
+         utils.mkdirp([os.path.dirname(module.module_dir), module.container_dir])
+        else: 
+         utils.mkdirp([module.module_dir, module.container_dir])
 
         # Add a .version file to indicate the level of versioning
         self.versionfile.write(
@@ -470,12 +473,17 @@ class ModuleBase(BaseClient):
 
         # Get the template based on the module and container type
         template = self.template.load(self.templatefile)
-        module_path = os.path.join(module.module_dir, self.modulefile)
+        if self.settings.version_naming:
+            self._modulefile = name.split(":")[1]             
+            module._module_dir=os.path.dirname(module.module_dir)
+            module_path=os.path.join(os.path.dirname(module.module_dir),self._modulefile)            
+        else:  
+           module_path = os.path.join(module.module_dir, self.modulefile)
 
         # Install the container
         # This could be simplified to take the module
         self.container.install(module_path, template, module, kwargs.get("features"))
-
+                
         # If the container tech does not need storage, clean up
         if not os.listdir(module.container_dir):
             utils.remove_to_base(module.container_dir, self.container_base)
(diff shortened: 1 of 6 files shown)


shpc/main/modules/base.py
```diff
@@ -450,15 +450,18 @@ def install(
         "force" is currently not used.
         """
         # Create a new module
-        module = kwargs.get("module") or self.get_module(
+         module = kwargs.get("module") or self.get_module(
```
vsoch, line 453: This is going to fail because the spacing is indented once.

shpc/main/modules/base.py
```diff
@@ -450,15 +450,18 @@ def install(
...
             name, container_image=container_image, keep_path=keep_path
         )
 
         # We always load overrides for an install
         module.load_override_file()
 
         # Create the module and container directory
-        utils.mkdirp([module.module_dir, module.container_dir])
+        if self.settings.version_naming:
```
vsoch, line 461: Please run black on these files or pre-commit for consistent spacing.

Amjadhpc: Thanks, i have ran the black command for consistent spacing

vsoch:
Looks like there are issues with the updated runner - let me see what I can do in another PR.

vsoch:
@Amjadhpc could you please add a test that enables the new setting and ensures it produces the right output? And @muffato I'm interested in your feedback here since you are correct views could be an alternative solution.

vsoch:
I think you probably want to try:

```python
client.settings.set("version_naming", True)
assert client.settings.get("version_naming") == True
```

shpc/tests/test_version_naming.py
```diff
@@ -69,7 +69,7 @@ def test_disabled_version_naming_install(
     client = init_client(str(tmp_path), module_sys, container_tech)
 
     client.settings.set("version_naming", False)
-    assert client.settings.set("version_naming", False) is False
+    assert client.settings.set("version_naming") == False
```
vsoch, line 72: You need to first `set` it, and then use `get` to check the setting is what you set it to.

shpc/tests/test_version_naming.py
```diff
@@ -69,7 +69,7 @@ def test_disabled_version_naming_install(
     client = init_client(str(tmp_path), module_sys, container_tech)
 
     client.settings.set("version_naming", False)
-    assert client.settings.set("version_naming") == False
+    assert client.settings.get("version_naming") == False
```
vsoch, line 72: You had one more on line 46.

vsoch:
Nicely done! You'll want to (as a final step) bump the version in `shpc/version.py` and add a note to the CHANGELOG.md. @marcodelapierre and @muffato do you have any comments for the review?

shpc/main/modules/base.py
```diff
@@ -470,7 +473,14 @@ def install(
 
         # Get the template based on the module and container type
         template = self.template.load(self.templatefile)
-        module_path = os.path.join(module.module_dir, self.modulefile)
+        if self.settings.version_naming:
+            self._modulefile = name.split(":")[1]
```
vsoch, line 477: What is `self._modulefile` used for? I only see it in this small set of statements. It might make sense to just use `name.split(":")[1]` in the one place it is needed and not save a private class variable.

marcodelapierre:
Hi, it is always great to see contributions coming from users!

As per @muffato 's comment in #678 ,
I would like to know whether module views do not fit your usage scenario and in which aspects.  
In fact, the idea of views was to have a solid default structure for the default module tree, and then offer ways to customise the module tree in a user defined view. As @muffato mentions in the original issue, customising the naming of modules is one of the views' features.  

I think it would be good to know whether Views can work in your user case or not, to try and not duplicate functionalities if not needed.  Either way, discussing these aspects with users of the software has a great deal of added value, in making sure the software is useful to the community, both in terms of functionality and documentation on how to use them.

My 2 cents :)

vsoch:
I agree it's not good to duplicate functionality, but in that views adds complexity (and this change makes it possible to generate the default with versions) I'm wondering if this isn't a good addition regardless.

shpc/settings.yml
```diff
@@ -68,6 +68,9 @@ singularity_shell: /bin/sh
 podman_shell: /bin/sh
 docker_shell: /bin/sh
 
+#Experimental version naming tag so that modulefile is changed to version number . Default value is false. Issue number 26
```
marcodelapierre, line 71: I suggest having this right after `default_version` option, line 39, as they are related

marcodelapierre:
Good point @vsoch  :)   

I will not have the chance to test this PR.  

One comment I have is: is the PR been designed and tested while taking into account the implementation and options of the `default_version` setting?  I bet there may be some specific requirements/adjustments coming out of it.

Amjadhpc:
> Nicely done! You'll want to (as a final step) bump the version in `shpc/version.py` and add a note to the CHANGELOG.md. @marcodelapierre and @muffato do you have any comments for the review?

The version in shpc/version.py is 
__version__ = "0.1.29"

You want me to bump it to "0.1.30"?
But there is no Release for 0.1.29, the last release was 0.1.28?

shpc/main/modules/base.py
```diff
@@ -470,7 +473,13 @@ def install(
 
         # Get the template based on the module and container type
         template = self.template.load(self.templatefile)
-        module_path = os.path.join(module.module_dir, self.modulefile)
+        if self.settings.version_naming:
+            module._module_dir = os.path.dirname(module.module_dir)
```
vsoch, line 477: Can you please walk me through why we need a private class variable here?

Amjadhpc: Sorry, forgot to clean it up, its not needed and have been fixed in new commit

vsoch:
@Amjadhpc you'll need to go up to 0.1.31 now - we had another PR merge in this morning.

Amjadhpc:
Looks like it has reached docker pull rate limit 

WARNING: Couldn't use cached digest for registry: GET https://index.docker.io/v2/library/python/manifests/sha256:7dd8962ad2a63403428d652a64d814a5002f1386379355edf5970e40557fe4e6: TOOMANYREQUESTS: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
WARNING: Falling back to direct digest.
FATAL:   docker://python@sha256:7dd8962ad2a63403428d652a64d814a5002f1386379355edf5970e40557fe4e6: GET https://index.docker.io/v2/library/python/manifests/sha256:7dd8962ad2a63403428d652a64d814a5002f1386379355edf5970e40557fe4e6: TOOMANYREQUESTS: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

vsoch:
@Amjadhpc it's an ephemeral error, OK to run again. I'm not going to put credentials in there just for these rare cases.

vsoch (approved):
Nice work @Amjadhpc !

PR #682, opened 2025-01-11
Make "module-info shell" Tcl test compatible with Lmod<=8.7.55

xdelaruelle:
Up to Lmod 8.7.55, the Tcl command "module-info shell" only returns the current shell name and cannot test if a given string corresponds this current shell.

A patch is now merged on Lmod, so newer version (8.7.56+) will cope with the "module-info shell bash" syntax.

But in the mean time, I propose here to use a test Tcl syntax that will be compatible with both Lmod and Environment Modules: "[module-info shell] eq {bash}".

This change may only be of help for people willing to use the Tcl modules produced by SHPC with the Lmod module tool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 The versions coincide with releases on pip. Only major versions will be released as tags on Github.
 
 ## [0.0.x](https://github.com/singularityhub/singularity-hpc/tree/main) (0.0.x)
+ - Fix `module-info shell` Tcl test for Lmod<=8.7.55 (0.1.30)
  - Allow import of LooseVersion from packaging (bug) (0.1.29)
  - use quay.io api to list tags since does not conform to oci (0.1.28)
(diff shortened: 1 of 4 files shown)


shpc/settings.yml
```diff
@@ -5,13 +5,13 @@
...
 # please open an issue https://github.com/singularityhub/singularity-hpc
-module_sys: lmod
+module_sys: tcl
 
 # config editor
 config_editor: vim
 
 # set a default container technology (singularity or podman)
-container_tech: singularity
+container_tech: podman
```
vsoch, line 14: You'll want to leave the defaults here as is (the container_tech and remove updated_at). Other than that, just update the version in shpc/version.py and add a corresponding note to the changelog! It's after midnight and I'm off to sleep, but I'll finish the review in the morning. Thanks for fixing this!

xdelaruelle: My bad! the settings update is a mistake (it is too early in the morning on my side ^^).

I have fixed the commit and updated the version and changelog.

Have a good night!

vsoch (approved):

PR #689, opened 2025-04-09
Added an option to define the path of the `singularity` executable

muffato:
Hi @vsoch 

I'm trying to clean up my local branches. I found this one that has a small commit.
This is about adding an option to define where is `singularity`. It can be used on sites where `singularity` is not necessarily in the `$PATH` of all users or when it is not provided as a module.


docs/getting_started/user-guide.rst
```diff
@@ -203,6 +203,11 @@ variable replacement. A summary table of variables is included below, and then f
    * - singularity_module
      - if defined, add to module script to load this Singularity module first
      - null
+   * - podman_module
```
vsoch, line 206: Should podman module be here? The singularity path addition is fantastic - great for security too.

vsoch (commented):
ok, this looks good then! Please bump the version and add a note to the CHANGELOG.md and we should be good to merge.

muffato:
Done !

vsoch (approved):

PR #690, opened 2025-04-11
Need to modify the main wrapper script too !

muffato:
Sorry @vsoch but in #689  I forgot to update the most important of all scripts !


vsoch:
No worries! Can you bump this one more to 0.1.32? Turns out 0.1.31 was already on pypi so your previous change hasn't been released yet. The silver lining here is we can do that easily now.

muffato:
Ah right, I remember 0.1.31 was released and then reverted. Since it remains on Pypi, I thought we should document it somewhere. Take a look at how I put it back in the Changelog and introduced 0.1.32

vsoch (approved):
I think that looks perfect. Let's merge and deploy this one - for realsies this time :wink: 

Thanks for both the feature and the fix!

muffato:
Thank you !

PR #697, opened 2026-05-10
Feature: Add DockerImage classes for docker.io, nvcr.io, and quay.io

### LINKED ISSUE #480: Address dockerhub pull limit from clobbering files
I'm working on this one, but adding an issue here to bookmark what's going on!

Dockerhub has gotten really strict about pull limits, and artifacts of that have made their way into registry yamls like this golang one: https://github.com/singularityhub/shpc-registry/blob/main/golang/container.yaml

Ideally, I'm looking for a way to not run afoul of rate limiting in the auto update, but preventing garbage from making it to the yaml is first up for the fix. I'm starting by validating that it's a recognizable sha format.

georgiastuart:
This is intended to resolve https://github.com/singularityhub/shpc-registry/issues/480 and https://github.com/singularityhub/shpc-registry/issues/481 . The driving motivations are rate limiting from dockerhub and odd output from NGC containers at times. The updated registry files in https://github.com/singularityhub/shpc-registry/pull/483 use this singularity-hpc PR. This is a moderately sized PR, so I'm anticipating a couple rounds of iteration after the PR tests run.

## Subclass Design 

`DockerHubImage`, `QuayDockerImage` and `NGCImage` Subclass `DockerImage` and override the `tags` and `digest` methods. It's possible there's a more elegant way to simplify, but all these APIs have bespoke needs so I didn't try to unify further. All query metadata APIs rather than use `crane`. 

### NCG Compatibility 

Querying the NCG API requires an API token. These are free to get for this purpose, but I kept `ngcsdk` as an optional dependency. If `ngcsdk` and the environment variable `SHPC_NGC_API_KEY` are present, shpc will use `ngcsdk` to update. Otherwise, it will fall back to `crane` and the normal `DockerImage` class. 

## Added CLI arguments for `update`

While cleaning up some mangled `container.yaml` files, I added two CLI arguments to `update`: 

`--max-tags=N`: Overrides the default 5 tags to add
`--purge`: wipes the existing tags in the `container.yaml` file to start fresh.

### DIFF
--- a/shpc/main/container/update/docker.py
+++ b/shpc/main/container/update/docker.py
@@ -133,8 +133,7 @@ class DockerHubImage(DockerImage):
 
     def __init__(self, container_name):
         super().__init__(container_name)
-
-        container_name_array = container_name.lstrip("docker.io/").split("/")
+        container_name_array = container_name.replace("docker.io/", "", 1).split("/")
         if len(container_name_array) == 1:
             self.container_name = "library/%s" % container_name_array[0]
         elif len(container_name_array) == 2:
--- a/shpc/main/container/update/docker.py
+++ b/shpc/main/container/update/docker.py
@@ -3,6 +3,7 @@ __copyright__ = "Copyright 2021-2025, Vanessa Sochat"
 __license__ = "MPL 2.0"
 
 import re
+import time
 
 import requests
 
@@ -26,8 +27,17 @@ class DockerImage:
         Perform a get request, expecting status code 200.
         """
         response = requests.get(url)
+
+        # Try to retry once if rate limited
+        if response.status_code == 429:
+            retry_seconds = response.headers.get("Retry-After", "unknown")
+            logger.warning("Rate limit hit for %s, retrying after %s seconds" % (url, retry_seconds))
+            time.sleep(int(retry_seconds))
+            response = requests.get(url)
+
         if response.status_code != 200:
             logger.exit("Issue with request %s" % url)
+
         return response
 
     def tags(self):
@@ -87,6 +97,17 @@ class DockerImage:
         url = "%s/config/%s" % (self.apiroot, self.container_name)
         return self.get_request(url).json()
 
+class QuayDockerImage(DockerImage):
+
+    """
+    A thin client for getting metadata about an image on Quay.
+    """
+
+    def __init__(self, container_name):
+        super().__init__(container_name)
+        self.apiroot = "https://quay.io/api/v1/repository/%s/tag" % (
+            self.container_name.lstrip("quay.io/")
+        )
 class DockerHubImage(DockerImage):
 
     """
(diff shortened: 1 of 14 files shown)


georgiastuart:
Shockingly, the tests passed the first time. Ready for review @vsoch !

shpc/main/container/update/docker.py
```diff
@@ -86,3 +82,180 @@ def digest(self, tag):
...
+                repository,
+                page,
+                specific_tag,
+            )
+            response = self.get_request(url).json()
+            tags = response.get("tags", {})
+
+            if len(tags) == 0:
+                logger.error(
+                    f"The tag {tag} you provided is not known. Check that it and the container both exist."
```
vsoch: Suggested for here would be to raise the ValueError with the message currently given to logger.error.

shpc/main/container/update/docker.py
```diff
@@ -86,3 +82,180 @@ def digest(self, tag):
...
+                )
+                raise ValueError
+            new_tags = [x for x in tags if x.get("name")]
+            tags.extend(new_tags)
```
vsoch, line 149: Sanity check here we will not have duplicates in this list. And what about order? I think different registries may do alphabetical vs. by date. If the user asks for some specific number of tags, we likely want to give them latest.

shpc/main/container/update/docker.py
```diff
@@ -86,3 +82,180 @@ def digest(self, tag):
...
+
+        while True:
+            response = self.get_request(url)
+            tag_responses.extend(response.json()["results"])
+            url = response.json().get("next")
+            if url is None:
+                break
+        return tag_responses
+
+    def tags(self):
```
vsoch: You are going in the right direction by using DockerHubImage for NGC, and it looks like there is still some redundancy in these classes (e.g., the tags filter for the sbom and similar). The main difference tends to be in the query tag API function. Could we put more logic in the DockerHubImage base class that the others inherit from with the shared logic?

shpc/main/container/update/versions.py
```diff
@@ -35,18 +35,20 @@ def filter_versions(tags, filters=None, max_length=5):
     filters (list)   : an optional list of string filters
     max_length (int) : the max number to return (latest)
     """
+
+    filtered_tags = [x for x in tags]
```
vsoch, line 39: Sanity check - order and repeats?

shpc/version.py
```diff
@@ -2,7 +2,7 @@
 __copyright__ = "Copyright 2021-2025, Vanessa Sochat"
 __license__ = "MPL 2.0"
 
-__version__ = "0.1.33"
```
vsoch, line 5: Let's bump up to `0.2.0` for this merge. If you have other changes / PRs to get in we can encompass them with that release too.

vsoch (commented):
This looks great! A few comments, and then we should have a test for the registry classes. We will also want a strategy to clean up the current registry repository.

georgiastuart:
Thanks for the review!! I'll implement tests and the other suggestions (and check out the 2 line digest printing).

shpc/main/container/update/docker.py
georgiastuart: Refactored to minimize redundancy. All class specialization now occurs in `_query_tag_api`. I marked the `manifest` and `config` methods to raise `NotImplementedError` in the subclasses. It appears that they aren't currently used elsewhere. I intend to implement, but throwing the error in for now just in case!

shpc/main/container/update/docker.py
georgiastuart: Question about this - I mainly took inspiration from the previous `tag_quay` method where it pulls all the tags and feeds it into the filtering, value extraction, and sorting functions in the update main function. DockerHub does appear to reliably return tags in order of latest update. Given that this class can handle a rate limit, is it better to continue pulling all the tags, or to cap it at the first X tags? I'm guessing we could comfortably restrict it to 5 pages or so and still be able to get the N most recent tags for even tag-spammy projects.

shpc/main/container/update/docker.py
georgiastuart: Refactored error handling a bit, but it still seems messy. Going to take a fresh look at it tomorrow!

shpc/version.py
georgiastuart: Great, can do. I want to add gitlab.com and self hosted gitlab support in before merge. I'm also looking at a strategy to handle multiple "series" of tags, like "X.X.X-alpine" vs "X.X.X-trixie" or whatever.

georgiastuart:
To do before merge:

- [ ] Add gitlab.com / self hosted support
- [ ] Add tests for remote image types & error handling
- [ ] (If feasible) look at "tag series" refactor
- [ ] Force `tag: digest` yaml to print on one line

shpc/main/container/update/docker.py
vsoch: We should likely be conservative and cap at some X tags, especially if you are hitting a limit with requests.

shpc/version.py
vsoch: Good thinking. :+1:

muffato:
Hi @georgiastuart . I'd be happy to help with "Add gitlab.com / self hosted support". We started working on it with @vsoch in #585 but never got to complete it.

vsoch:
Thank you @muffato !

